PR #4, opened 2022-05-03
Add required query parameter to test optic-ci

bertramakers:
### Added

- [Added ...]

### Changed

- [Changed ...]

### Removed

- [Removed ...]

### Fixed

- [Fixed ...]

---

Ticket: https://jira.uitdatabank.be/...


bertramakers:
The optic-ci check is currently crashing on a cryptic error. I suspect it might be because it doesn't support JSON files (yet) but only YAML? Will check later.

PR #5, opened 2022-05-03
III-4667 Add docs for updating attendance mode

LucWollants:
### Added
- Added docs for updating attendance mode

---

Ticket: https://jira.uitdatabank.be/III-4667

### DIFF
--- a/projects/uitdatabank/models/event-attendanceMode.json
+++ b/projects/uitdatabank/models/event-attendanceMode.json
@@ -1,0 +1,10 @@
+{
+  "title": "event.attendanceMode",
+  "type": "string",
+  "description": "Indicates if the [event](./event.json) is happening online, offline, or both (mixed).",
+  "enum": [
+    "online",
+    "offline",
+    "mixed"
+  ]
+}
--- a/projects/uitdatabank/reference/entry.json
+++ b/projects/uitdatabank/reference/entry.json
@@ -760,7 +760,7 @@
           }
         },
         "operationId": "event-attendance-mode-put",
-        "description": "Updates the attendance mode of an event. There are three different attendance modes:\n- `offline`: the event takes places on a real location\n- `online`: the events takes places on a virtual location\n- `mixed`: the event takes places both on a real location and a virtual location\n\nWhen changing from attendance mode `online` to either `offline` or `mixed` it is important to pass the uri of the location. This location uri will be used to replace the virtual location with the existing real location.\n\nA virtual location is a special kind of location with UUID `00000000-0000-0000-0000-000000000000`",
+        "description": "Updates the attendance mode of an event. There are three different attendance modes:\n- `offline`: the event takes places on a real location\n- `online`: the events takes places on a virtual location\n- `mixed`: the event takes places both on a real location and a virtual location\n\nWhen changing from attendance mode online to either offline or mixed it is required to include the location property with the URI or UUID of the (physical) [location](/models/place.json) that the event is taking place at.",
         "security": [
           {
             "USER_ACCESS_TOKEN": []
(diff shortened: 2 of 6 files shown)


projects/uitdatabank/reference/entry.json
```diff
@@ -731,6 +731,69 @@
...
       }
     },
+    "/events/{eventId}/attendance-mode": {
+      "parameters": [
+        {
+          "$ref": "#/components/parameters/eventId"
+        }
+      ],
+      "put": {
+        "summary": "attendance-mode - update",
```
bertramakers, line 741:
```suggestion
        "summary": "attendanceMode - update",
```

For other operations I've also always used the property name instead of the path name. The reasoning being that an integrator would look how to update a specific property. The path names just happen to use dashes instead of camelCase because that's the convention for URLs.

projects/uitdatabank/models/event-attendanceMode.json
```diff
@@ -0,0 +1,10 @@
+{
+  "title": "event.attendanceMode",
+  "type": "string",
+  "description": "Indicates if the [event](./event.json) is happening online, offline, or both (mixed).",
+  "enum": [
+    "online",
+    "offline",
+    "mixed"
+  ]
+}
```
bertramakers, line 10: For completeness it would be best to also link this in the `toc.json` under the "Event" group where all event properties are currently listed: https://github.com/cultuurnet/apidocs/blob/main/projects/uitdatabank/toc.json#L45-L231

I'm going to double check with Corneel if this is still useful though. I initially added it for Sven before we had any endpoints documented and Stoplight still used their previous design, to make it easier for him to show our schema to some partners. But it might be redundant now since the properties can now all be clearly seen on the "Event" model itself.

projects/uitdatabank/reference/entry.json
```diff
@@ -731,6 +731,69 @@
...
+          },
+          "403": {
+            "$ref": "#/components/responses/Forbidden"
+          },
+          "404": {
+            "$ref": "#/components/responses/EventNotFound"
+          }
+        },
+        "operationId": "event-attendance-mode-put",
+        "description": "Updates the attendance mode of an event. There are three different attendance modes:\n- `offline`: the event takes places on a real location\n- `online`: the events takes places on a virtual location\n- `mixed`: the event takes places both on a real location and a virtual location\n\nWhen changing from attendance mode `online` to either `offline` or `mixed` it is important to pass the uri of the location. This location uri will be used to replace the virtual location with the existing real location.\n\nA virtual location is a special kind of location with UUID `00000000-0000-0000-0000-000000000000`",
```
bertramakers, line 763: > When changing from attendance mode online to either offline or mixed it is important to pass the uri of the location. This location uri will be used to replace the virtual location with the existing real location.
> 
> A virtual location is a special kind of location with UUID 00000000-0000-0000-0000-000000000000

I'd suggest to rephrase this to:

> When changing from attendance mode `online` to either `offline` or `mixed` it is required to include the `location` property with the URI or UUID of the (physical) [location](../models/place.json) that the event is taking place at.

projects/uitdatabank/models/event-attendanceMode.json
LucWollants: Should we also order this TOC better?

bertramakers: Hmm yes but I'll check with Corneel if it's actually still needed, I'll get back to you about it asap

bertramakers (approved):
Let's merge this like it is now since it all looks good, and we can add the link for the `toc.json` in another PR if needed or remove them all

PR #10, opened 2022-05-10
pubmpm-923 museumpass partner api

erwin1:
### Added

- museumpass partner api


bertramakers (approved):
Still missing an `errors.md`, and a good guide to explain the complete flow of a museum visit and for selling vouchers, but those can be added later.

bertramakers:
Marked the `/POST visits` and `GET /passes/{cardNumber}` as internal for now as they are not implemented yet, but we want to merge this and make the `POST /voucher-lists` visible for readers.
Still needs an `errors.md` but only when we release `/POST visits` (only endpoint to use custom error types at the moment).

PR #17, opened 2022-05-18
III-4732 Use `url/not-found` for all 404s

bertramakers:
### Added

- Added new custom `attendance-mode-not-supported` error type

### Changed

- Changed all 404 responses to use the same response with type `https://api.publiq.be/probs/url/not-found`

### Removed

- Removed references to old URL error types

---

Ticket: https://jira.uitdatabank.be/browse/III-4732


projects/uitdatabank/docs/errors.md
```diff
@@ -77,3 +77,13 @@ Example error response:
...
+
+## attendance-mode-not-supported
+
+*   **Complete type:** `https://api.publiq.be/probs/uitdatabank/attendance-mode-not-supported`
+*   **Title**: `Attendance mode not supported`
+*   **Status**: `400`
+
+The action you are trying to perform on an event cannot be done because its `attendanceMode` does not support this action.
+
+For example, you cannot update the location of an event with attendanceMode `online` to a physical location. Or you cannot set the location of an event with attendanceMode `mixed` or `offline` to the nil location (`https://io.uitdatabank.be/places/00000000-0000-0000-0000-000000000000`).
```
LucWollants, line 89: Everything was changed to `nil location` in the code, but inside documentation, I still used `online location` because that is clearer for the current context.

Dot we also use `nil location` in documentation?

LucWollants (approved):
I already approved, the `nil location` versus `online location` is just a tweak.

projects/uitdatabank/docs/errors.md
bertramakers: To me in a technical context `nil location` is clearer personally because nil = 0 and `00000000-0000-0000-0000-000000000000` is officially called the nil uuid. But maybe I'm biased because I happen to know that term for the uuid while the average integrator might not.

On the other hand I'm not sure that `00000000-0000-0000-0000-000000000000` = `online location` would be clear at first glance without reading about online events and attendance mode first. 🤔

LucWollants: Ok, I will make sure to use `nil location` inside the code and inside documentation.

bertramakers: As discussed on Slack we'll keep calling it nil location in the docs so this does not need to be changed. But we'll probably need to change some other occurences of "online location" in the docs to "nil location". And also clarify what it means in the terminology page.

PR #20, opened 2022-05-25
add priceinfo put request for Events & Places

JonasVHG:
### Added

- put request for `PriceInfo` on events
- put request for `PriceInfo` on places

---

Ticket: https://jira.uitdatabank.be/browse/III-4268

### DIFF
--- a/projects/uitdatabank/models/event-priceInfo-put.json
+++ b/projects/uitdatabank/models/event-priceInfo-put.json
@@ -1,0 +1,12 @@
+{
+  "title": "event.priceInfo.put",
+  "type": "object",
+  "properties": {
+    "onlineUrl": {
+      "$ref": "./event-priceInfo.json"
+    }
+  },
+  "required": [
+    "priceInfo"
+  ]
+}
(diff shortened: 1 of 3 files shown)


projects/uitdatabank/models/event-priceInfo-put.json
```diff
@@ -0,0 +1,12 @@
+{
```
LucWollants, line 1: I wonder if these files are required and even if they are correct. If you check the old documentation https://documentatie.uitdatabank.be/content/json-ld/latest/events/event-price-info.html and the acceptance tests https://github.com/cultuurnet/udb3-acceptance-tests/blob/main/features/event/price-info.feature#L15 then the extra property `priceInfo` is not present on the request body.

That also explains why the current validation is working with `event-priceInfo.json`. Only an extra model for `PUT` is needed when an additional property is present.

LucWollants: If this is not required, make sure to also update or even remove the example.

LucWollants (approved):

PR #22, opened 2022-05-30
III-4658 Remove readOnly properties from request examples

bertramakers:
### Fixed

- Fixed organizer, place and event examples so that readOnly properties are only shown in the GET request and page of the organizer/place/event model, not in POST/PUT requests.

---

Ticket: https://jira.uitdatabank.be/browse/III-4658


bertramakers:
Failing check is due to https://www.uitinvlaanderen.be (referenced in the Markdown docs) being down, causing the dead link checker to fail.

LucWollants (approved):
Looks good, I also checked https://publiq.stoplight.io/

PR #23, opened 2022-06-01
III-4761 + III-4762 Event/place example tweaks

bertramakers:
### Added

- Added missing `videos` to event create/update examples

### Changed

- Grouped `status` and `bookingAvailability` together in event and place update examples because they have a high cohesion functionally

### Removed

- Removed `status` and `bookingAvailability` from event/place create examples because they are usually not used when creating events/places, but only set later (see ticket description)
- Removed `availableTo` from event/place create & update examples, because it's a read-only property that gets derived from the calendar info. So it is not useful to include it in create/update requests.

---

Ticket: https://jira.uitdatabank.be/browse/III-4761 + https://jira.uitdatabank.be/browse/III-4762

### DIFF
--- a/projects/uitdatabank/reference/entry.json
+++ b/projects/uitdatabank/reference/entry.json
@@ -230,6 +230,15 @@
                         "inLanguage": "nl"
                       }
                     ],
+                    "videos": [
+                      {
+                        "id": "b504cf44-9ab8-4641-9934-38d1cc67242c",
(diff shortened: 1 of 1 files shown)


bertramakers:
The linting is failing on a "dead link" check on https://github.com/cultuurnet/apidocs/actions
When I visit that link I get a 500, and I also had problems creating this PR. So I think GitHub is having some temporary issues. I'll re-run the job when it's fixed.

PR #32, opened 2022-06-29
Uitpas/ups 3610 rewards details

erwin1:
### Changed

- redeemConstraint.volume is integer instead of string

### Removed

- reward.redeemType: until the use cases are clear

Ticket: https://jira.uitdatabank.be/browse/UPS-3610

### DIFF
--- a/projects/uitpas/reference/uitpas.json
+++ b/projects/uitpas/reference/uitpas.json
@@ -3136,7 +3136,7 @@
                           "points": 10,
                           "forKids": false,
                           "sport": false,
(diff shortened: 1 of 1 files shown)


bertramakers (approved):
Is the `"labels": ""` a realistic example? Can that property be an empty string? Or will it just not be in the object if it's empty?
Aside from that this looks good!

erwin1:
> Is the "labels": "" a realistic example?

apparently most rewards have indeed an empty labels field

PR #39, opened 2022-07-05
PUBMPM-932 museumpassmusees partner api

erwin1:
### Added

- endpoints for external card visits

---

Ticket: PUBMPM-932


projects/museumpassmusees/reference/partner-api.json
```diff
@@ -430,9 +430,188 @@
...
+                      "description": "Visit information for the external card.",
+                      "properties": {
+                        "allowed": {
+                          "type": "boolean",
+                          "description": "Indicates whether a visit with the external card is allowed or not."
+                        },
+                        "reason": {
+                          "type": "string",
+                          "enum": [
+                            "NO_SOCIAL_STATUS"
```
bertramakers, line 459: Are `EXPIRED` and `BLOCKED` not possible reasons in the case of UiTPAS and/or A-Kaart?

projects/museumpassmusees/reference/partner-api.json
```diff
@@ -430,9 +430,188 @@
...
+            "content": {
+              "application/problem+json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            }
+          },
+          "503": {
+            "description": "Service Unavailable. Possible error types:\n\n* https://api.publiq.be/probs/museumpass/external-service-unavailable\n\nThe `Retry-After` HTTP Header might include a hint when a retry of this request can be attempted.\n\n",
```
bertramakers, line 491: > Possible error types:\n\n* https://api.publiq.be/probs/museumpass/external-service-unavailable\n\n

For self-explanatory status codes you MAY also use `about:blank` as the type according to RFC7807. For example 403, 404 or 500. The reason that we use a custom type for 4XX errors even when not needed is that we can make it link to documentation to provide some more info how to solve it. But for 5XX errors I'm not sure that that is necessary since a client can not solve it on their end. So for 5XX errors I think `about:blank` is fine as type, and then you don't need to document the error type(s) here either.

projects/museumpassmusees/reference/partner-api.json
```diff
@@ -430,9 +430,188 @@
...
+            "content": {
+              "application/problem+json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            }
+          }
+        },
+        "description": "Registers a visit of an axternal card at your museum (determined by the client access token used).\n\nTo check if this card is allowed to visit, look up their information using [`GET /external-cards/{type}/{identification}`](./partner-api.json/paths/~1get-external-cards-type-identification~1{type}~1{identification}/get). The response to that request will indicate if the passholder may visit your museum.\n\nIf the external card may not visit your museum and you still try to register a visit for them, the server will return a `403 Forbidden` error response.\n\n## Permissions\n\nThe caller of this request must have the `external card visit registrar` role.",
```
bertramakers, line 574: > look up their information using [`GET /external-cards/{type}/{identification}`](./partner-api.json/paths/~1get-external-cards-type-identification~1{type}~1{identification}/get)

When I try the link on https://docs.publiq.be/docs/museumpassmusees/branches/museumpassmusees%2FPUBMPM-932-partner-api/2d1c386cb2642-register-a-museum-visit-for-an-external-card it does not work. But I don't see the issue immediately either. Maybe try with an absolute path? (So prefixing it with `/reference/` instead of `./`)

bertramakers, line 574: Since it's uncommon for POST requests to not have a body, it might be good to mention explicitly in the description that it a body is not required. Otherwise it might seem like we forgot to document it.

projects/museumpassmusees/reference/partner-api.json
erwin1: From a functional perspective MPM only has to check for 'social status' - the question is: if we would need to extend this later, is it better to include EXPIRED and BLOCKED now (but ignore them in the implementation) or add them when needed. (extra enum values could be seen as a breaking change)

projects/museumpassmusees/reference/partner-api.json
bertramakers: @erwin1 I tried it again after your change to prefix it with `/reference` but that didn't fix it either. I noticed that the actual path was incorrect so I fixed it in [82f233d](https://github.com/cultuurnet/apidocs/pull/39/commits/82f233d66af72a3df39283a98932f5ea6519d636).

One way to easily get the correct URL is by right-clicking the endpoint in Studio and clicking "Copy relative path". Then you can paste wherever you need it, and add `/get` or `/post` or ... to the URL depending on the method you want to link to.

bertramakers (approved):

projects/museumpassmusees/reference/partner-api.json
bertramakers: As discussed on Slack the `reason` is just left out for now, since it doesn't give any extra info if it can only be `NO_SOCIAL_STATUS` for now.

bertramakers (approved):

PR #44, opened 2022-07-06
Uitdatabank/search-api-guides-case-sensitivity

corneelwille:
### Added
- Documentation page for case sensitivity
- Link to case sensitivity docs to sidebar

### DIFF
--- a/projects/uitdatabank/docs/search-api/case-sensitivity.md
+++ b/projects/uitdatabank/docs/search-api/case-sensitivity.md
@@ -1,0 +1,99 @@
+# Case sensitivity
+
+Unless noted otherwise:
+-   All parameters are `case-sensitive`
+-   All values are `case-insensitive`
+
+## Parameters are case-sensitive
+
+Every parameter should be written in `camelCase`: seperation of words is done with a single capitalized letter, and the first word starting with either case.
+
+### Examples
+
+#### Using the `bookingAvailability` parameter correctly (with camelCase)
+
+**request**
+```
+GET /offers/?q=bookingAvailability:unavailable
+```
+
+**response**
+
+Results are returned ✅
+```js
+{
+   "@context":"http://www.w3.org/ns/hydra/context.jsonld",
+   "@type":"PagedCollection",
+   "itemsPerPage":30,
+   "totalItems":157,
+   "member":[ ],
+   ...
+}
+```
+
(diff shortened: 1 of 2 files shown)


bertramakers (commented):
Same as with the guide about escaping, this seems very specific to the `q` parameter so better suited as part of that guide I think?

corneelwille:
> 

In the documentation both examples are with a `q` parameter, but I think this applies to url parameters too:

**case sensitivity of the parameters**
- `bookingAvailability=Unavailable` returns results
- `bookingavailability=Unavailable` returns a 404 error `"unknown query parameter(s): bookingavailability"`

**case insensitivity of the values**
`labels=UiTPAS` returns the exact same results als `labels=uitpas`

So maybe we should just clarify and replace one of the current examples with a `url parameter` example?

bertramakers:
> > 
> 
> In the documentation both examples are with a `q` parameter, but I think this applies to url parameters too:
> 
> **case sensitivity of the parameters**
> 
>     * `bookingAvailability=Unavailable` returns results
> 
>     * `bookingavailability=Unavailable` returns a 404 error `"unknown query parameter(s): bookingavailability"`
> 
> 
> **case insensitivity of the values** `labels=UiTPAS` returns the exact same results als `labels=uitpas`
> 
> So maybe we should just clarify and replace one of the current examples with a `url parameter` example?

URLs and their query parameters are (almost) always case-sensitive though: https://stackoverflow.com/questions/24699643/are-query-string-keys-case-sensitive (only the host, i.e. the domain name, is case-insensitive). So that is not really special to Search API, and the parameters will be documented with the correct casing in the OpenAPI file. So I don't think that requires extra clarification? We don't mention that anywhere else for other APIs' query parameters.

For the `q` I think it's worth mentioning because someone could assume that the fields in the Lucene query syntax are case-insensitive, because it's not something that is commonly known by a lot of developers.

bertramakers:
Also because the `q` parameter is so complex for people who are not familiar with it, I'd even suggest to have multiple guides for it and group them together in the navigation. For example:

**Search API**
- ... other guides ...
- Advanced queries
  - Syntax
  - Fields
  - Case sensitivity
  - Escaping
  - etc ...

If you want I can help with those when I'm done with the OpenAPI file?

bertramakers:
@corneelwille do we close this one and put the relevant info in the advanced queries guide?

PR #45, opened 2022-07-07
Uitdatabank/search-api-guides-escaping

corneelwille:
### Added
- Escaping documentation page
- Link to escaping docs to sidebar

### DIFF
--- a/projects/uitdatabank/docs/search-api/escaping.md
+++ b/projects/uitdatabank/docs/search-api/escaping.md
@@ -1,0 +1,22 @@
+# Escaping
+
+Special characters like `:`, `"` or `*` should always be escaped.
+
+## Examples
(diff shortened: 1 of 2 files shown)


bertramakers (commented):
While I think this info is super useful (as I often forget this myself sadly), this guide feels super specific to the `q` parameter. And we'll need a guide about that parameter in itself as it is way too complicated to document in the OpenAPI file because it's a whole query language basically. So maybe this escaping info would best be included directly in that guide instead of a separate guide? We can of course copy-paste it from here to the guide for the `q` parameter, this content is definitely not lost!

corneelwille:
I'm moving this documentation to the advanced query docs (https://github.com/cultuurnet/apidocs/pull/65/)

PR #33, opened 2022-06-30
UiTdatabank - Search API documentation

corneelwille:
### Added
- Sorting guide

PR #50, opened 2022-07-25
Search API OpenAPI reference

bertramakers:
### Added

- Search API v3 OpenAPI file


bertramakers:
Failing check is because uitinvlaanderen.be is down and the dead linker returns an error for that

bertramakers:
@JonasVHG @LucWollants I'm going to go over this myself again in a day or two to look at it with a fresh pair of eyes to see that I haven't forgotten something important. You can already go over it to give feedback if you have some time. Note that descriptions can always be fine-tuned later. We'll need to go over those again anyway to add the actual links to the guides. Most importantly right now is that all requests, URL parameters and responses are documented.

Currently the PR points to `main` but I would merge this into https://github.com/cultuurnet/apidocs/pull/33 when ready so we have all the guides and OpenAPI specs in one place, and we can merge that when all SAPI3 documentation is ready for release. But to do that I need to merge that branch into this one first.

projects/uitdatabank/reference/search.json
```diff
@@ -0,0 +1,1674 @@
...
+            "$ref": "#/components/parameters/price"
+          },
+          {
+            "$ref": "#/components/parameters/minPrice"
+          },
+          {
+            "$ref": "#/components/parameters/maxPrice"
+          },
+          {
+            "$ref": "#/components/parameters/sortScore"
```
LucWollants, line 245: I think it will be difficult to group the sort parameters as another heading like `Query Parameters` but for example `Sort Parameters`?

bertramakers: No that's sadly not possible, OpenAPI only supports the following parameters:

- path parameters, such as `/users/{id}`
- query parameters, such as `/users?role=admin`
- header parameters, such as `X-MyHeader: Value`
- cookie parameters, which are passed in the Cookie header, such as `Cookie: debug=0; csrftoken=BUSe35dohU3O1MZvDCU`

(See https://swagger.io/docs/specification/describing-parameters/)

The "query" in that case corresponds to the URL query component as described in https://datatracker.ietf.org/doc/html/rfc3986#section-3.4 Basically everything after `?` is a query parameter, regardless what it's used for. So it does not mean "query" in the sense of "querying search results".

Stoplight then lists these different kinds of parameters under various titles, for example Headers, Query parameters, etc...

LucWollants: Thanks for clarifying this!

LucWollants (approved):
I did browse through the online documentation which seemed logical (I did not go into detail of every parameter and did not verify every example and/or validation rule)

bertramakers:
Okay I will merge this into #33 where it will get merged with the Search API guides that are done, but it won't be visible on "Stable" / "Unreleased" yet so we can still make changes to it when necessary. I'll also ask Corneel to go over it when he's back.

PR #55, opened 2022-08-10
API-89 Document overview of supported authentication methods per API

bertramakers:
### Added

- Added an overview of the supported authentication methods per API, to make it more clear at a glance which authentication method is relevant for an integrator who's reading the authentication docs.

### Changed

- Slightly optimized some wording of the sentences in the touched file.

---

Ticket: https://jira.uitdatabank.be/browse/API-89

### DIFF
--- a/projects/authentication/docs/methods.md
+++ b/projects/authentication/docs/methods.md
@@ -12,7 +12,6 @@ There are 3 possible scenarios for an endpoint:
 2.  The endpoint requires **[client identification](#client-identification)**. You will only need to include your client id in the request and you don't need a token. For example Search API.
 3.  The endpoint requires a **[token](#tokens)**. You can pick whatever token type is best suited to your situation (almost every endpoint will support both types). See their respective documentation for more info.
 
-> An endpoint that requires authentication will require either client identification (#2) **or** a token (#3), never both.
 The following table provides a summary of the supported authentication methods on our APIs. Note that APIs may have a few exceptions on specific endpoints. APIs that require no authentication at all, like [UiTdatabank Taxonomy API v3](https://docs.publiq.be/docs/uitdatabank/9b63d9296d981-taxonomy-api), are not included in this table.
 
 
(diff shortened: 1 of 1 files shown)


bertramakers:
Some changes that will need to be made in the future:

- UiTPAS API v4 will support client identification on the new `GET /rewards` endpoint (not released yet)
- Search API v3 will support client access tokens & user access tokens, next to client identification, for integrators that already have a token anyway to authenticate on another API

projects/authentication/docs/methods.md
```diff
@@ -4,31 +4,45 @@ Each publiq API supports one of three authentication methods, depending on the s
...
 
-You can find the authentication method(s) that an endpoint supports in its own documentation.
+APIs that require no authentication at all, like [UiTdatabank Taxonomy API v3](https://docs.publiq.be/docs/uitdatabank/9b63d9296d981-taxonomy-api), are not included in this table.
 
-Below you can find a short overview of how each authentication method works.
+
+API | Client identification | Client access tokens | User access tokens
+---------|----------|---------|---------
+ UiTdatabank Search API v3 | ✅ | No | No
+ [UiTdatabank Entry API v3](https://docs.publiq.be/docs/uitdatabank/5c4fcee8036e0-entry-api) | No | ✅ | ✅
```
corneelwille, line 23: Nit: maybe we should clean up the slug and link to https://docs.publiq.be/docs/uitdatabank/entry-api?

corneelwille (approved):
Very clear to me 👌

projects/authentication/docs/methods.md
bertramakers: Good point. I suggest that we make another PR later to clean up the slugs of existing pages now that that is possible, and we then also clean up the existing links. Then we don't end up with a mix of ugly and pretty slugs. If okay with you?

bertramakers: To keep things moving I'm going to merge this since you approved anyway. I can always make another PR to change this specific slug if you want, or we can do it as part of a larger cleanup a bit later.

PR #59, opened 2022-08-18
API-80 Document token flows with diagrams / steps

bertramakers:
### Added

- Added a diagram and steps how the client access token flow works. There was already an example.
- Added a diagram, steps and an example for how the user access token flow works (both for backend apps, as well as native/mobile/frontend apps)

### Changed

- Made the info about caching & expiration and decoding tokens consistent between the two token guides

---

Ticket: https://jira.uitdatabank.be/browse/API-80

Notes: 

- I will add info about refresh tokens (for user access tokens) in another PR later. See #60 
- I will also document the extra parameters like `prompt=login` for the `/authorize` redirect in another PR later.
- The text in the diagrams may be a bit blurry in some cases. This is due to a bug in the diagram component that I cannot fix, we need to wait until it's fixed in the library and then updated in Stoplight.

Preview:

- Client access tokens: https://docs.publiq.be/docs/authentication/branches/authentication%2FAPI-80-token-flows/4706c3721316a-client-access-token
- User access tokens: https://docs.publiq.be/docs/authentication/branches/authentication%2FAPI-80-token-flows/a6226e36a7627-user-access-token


projects/authentication/docs/client-access-token.md
```diff
@@ -16,30 +16,27 @@ If your browser or native application cannot work with user logins via publiq's
...
-> If you use or store your client secret or client access token in a frontend application they can be **stolen** and **abused** and your client will get blocked!
->
-> If you need to do API calls from a frontend application, use [client identification](./client-identification.md) or a [user access token](./user-access-token.md) depending on which methods the API endpoint accepts.
-
-> ##### OAuth2
->
-> Client access tokens are requested using the standardized [OAuth 2.0 Client Credentials Grant](https://oauth.net/2/grant-types/client-credentials/). If you are familiar with this flow, you can skip most of the example flow below. Do check the info about the required `audience` property though.
+## How it works
+
+```mermaid
```
corneelwille, line 21: The diagram is very clear (not blurry at all as mentioned in the description of your pull request) 👍

corneelwille:
I think it can be useful to repeat in (or just after) the introduction text on which APIs the client access tokens can be used. The overview you've made (https://docs.publiq.be/docs/authentication/branches/authentication%2FAPI-80-token-flows/6c0893b8e313d-overview) is very clear but it is possible that an integrator doesn't read that page at all and thus misses that crucial part of information when reading the client access token docs.

corneelwille (approved):

bertramakers:
> I think it can be useful to repeat in (or just after) the introduction text on which APIs the client access tokens can be used. The overview you've made (https://docs.publiq.be/docs/authentication/branches/authentication%2FAPI-80-token-flows/6c0893b8e313d-overview) is very clear but it is possible that an integrator doesn't read that page at all and thus misses that crucial part of information when reading the client access token docs.

I thought about that as well and created a ticket to add a text at the top of every page that links to the overview to get an overview of all possible methods and the APIs that they are usable on: https://jira.uitdatabank.be/browse/API-92

Otherwise we have to maintain documentation of what method is usable on which API in 3 places:
- The overview page
- The pages for the specific methods
- The docs of the APIs themselves

Would this solution also be okay for you? Or do we need the APIs on the pages for the specific methods?

projects/authentication/docs/client-access-token.md
bertramakers: For me the labels of the arrows are a bit "blurry" around the edges of the text.

Screen Shot 2022-08-19 at 10 57 05

They look like they're bold, but in fact their font weight is 400 (normal, not bold), but they have a "stroke". That is a feature in SVGs to add a border of a different color around objects and text. It seems to be that stroke makes the text look "bold" but also not as crispy as if it would just use font weight 700.

(With the small screenshot it's maybe not super visible here)

For example if I disable the stroke in my web inspector it already looks a lot better:

Screen Shot 2022-08-19 at 10 57 01

And when I then add font-weight 700 it looks bold again, but the edge is noticeable less blurry than in my first screenshot:

Screen Shot 2022-08-19 at 10 58 04

bertramakers: Anyway I found an issue in the mermaid JS library about this so I assume it will be fixed in the future. And it's still readable, just not ideal.

bertramakers:
Going to merge this already since the suggestion to link to the overview / mention all the supported APIs is probably better for another PR anyway (this one focuses on explaining the flows)

PR #65, opened 2022-08-24
Uitdatabank/search api guides advanced queries

corneelwille:
### Added
Documentation about:
* all advanced query parameters that can be used on the `/events`, `/places`, `/offers` and `/organizers` endpoint.
* the usage of boolean operators
* the usage of round brackets

For every parameter I followed the same structure:
- an introduction with some context and information about what the parameter can be used for
- applicable endpoints
- possible values
- example(s)

I think this structure makes the documentation easier to digest / skim / read. The structure gives some guidance about which information can be found where for each parameter:

Schermafbeelding 2022-08-24 om 15 23 51


projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1020 @@
...
+**Example**
+
+Retrieve all events that took place in the year 2020:
+```
+GET /events/?q=dateRange:[2020-01-01T00\:00\:00%2B01\:00 TO 2021-12-31T23\:59\:59%2B01\:00]&availableTo=*&availableFrom=*
+```
+
+### \_exists_
+
+With the `_exists_` parameter you can search for the documents that have a specific property in their JSON body.
```
bertramakers, line 303: Technically it returns documents that have the given field name in their index. Which is a bit different, because we do not index all JSON fields, or we index them with a slightly different name.

For example all events have a `sameAs` property with a link to UiTinVlaanderen, but we don't index that so `_exists_:sameAs` will return 0 results.

And the reverse is also true: `_exists_:address` will return places as well as events, while events never have an `address` property in their JSON. Because we index an event's location's address as `address`, so you can search on `address` in both events and places at the same time

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1020 @@
...
+
+**Applicable on endpoints**
+
+`/events` `/places` `/offers` `/organizers`
+
+**Possible values**
+
+Any property name from the UiTdatabank JSON scheme.
```
bertramakers, line 311: (See other comment) Not all properties from the UiTdatabank JSON schemas, and some with different names. It can only be the field names listed in this document basically

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1020 @@
...
+
+Retrieve events that match a specific uuid with the `id` parameter. It is possible to search by an event id, place id or organizer id.
+
+A specific id can be found by
+-   looking for the value for the `id` property
+```js
+{
+   "id": "75573a64-ddc8-4fd0-8b07-d258939dd74f"
+}
+```
```
bertramakers, line 331: In what scenario is this? Our events, places and organizers currently do not have an `id` property.

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1020 @@
...
+
+**Applicable on endpoints**
+
+`/events` `/places` `/offers` `/organizers`
+
+**Possible values**
+
+Any uuid of an event, place or organizer.
```
bertramakers, line 346: As with `creator` I think we shouldn't focus too much on the format (uuid) here, as we already have some ids that are not conform that format and we don't want to encourage integrators to depend on it

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1020 @@
...
+```
+
+Retrieve all organizers that have 5 to 10 images:
+```
+GET /organizers/?q=imagesCount:[5 TO 10]
+```
+
+### labels
+
+Search for documents that have a specific label with the `labels` parameter.
```
bertramakers, line 383:
```suggestion
Search for documents that have a specific label (or hidden label) with the `labels` parameter.
```

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1020 @@
...
+
+`/events` `/places` `/offers` `/organizers`
+
+**Possible values**
+
+A label name (string).
+
+**Example**
+
+Retrieve all events with an UiTPAS label:
```
bertramakers, line 395: `Paspartoe` is also an UiTPAS label and it will not match this query: https://io.uitdatabank.be/uitpas/labels. It's better to use the `uitpas` URL parameter to find UiTPAS events. 

So maybe use another example here that doesn't imply that this is the way to find UiTPAS events.

bertramakers (commented):
I haven't reviewed the whole PR yet, but I'm sharing my comments so far so you can already check those out and process them. I'll then go over the whole guide again later

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1020 @@
...
+
+Consequently we need to link both OR-relations with an AND operator:
+```
+/events/?q=address.\*.postalCode:(9000 OR 3000) AND terms.id:(0.0.0.0.0 OR 0.50.4.0.0)
+```
+Now the query will return concerts and expositions that either happen in Ghent or Leuven. Without the usage of round brackets you would get incomplete or incorrect results.
+
+## Escaping
+
+Special characters like `:`, `"` or `*` should always be escaped.
```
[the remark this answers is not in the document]
bertramakers: Also: Try to use `must` instead of `should` if something is required. In technical documents "should" is often used as a suggestion, and "must" for requirements.

projects/uitdatabank/docs/search-api/advanced-queries.md
bertramakers: If it helps to visualize it, you can find the properties that we index for events here: https://github.com/cultuurnet/udb3-search-service/blob/main/src/ElasticSearch/Operations/json/mapping_event.json

So what we do is we take the UDB JSON, then transform it to a new JSON document that has all or some of the properties listed in this mapping file. The mapping file is then used to tell Elasticsearch how to analyze specific fields. (Not how to transform the UDB JSON to the JSON documents in Elasticsearch!)

Basically what `_exists_` then does is check if a JSON document in Elasticsearch has one of these properties from the mapping file.

And it's also these field names that you can filter on in the advanced queries, if their analyzer supports it. (For example you cannot filter on `geo` or `geo_point` in the advanced queries because the `geo_shape` and `geo_point` field types are not supported in "query string queries" (= Elasticsearch name for what we call "advanced queries")

Here are the ones for places and organizers:

- https://github.com/cultuurnet/udb3-search-service/blob/main/src/ElasticSearch/Operations/json/mapping_place.json
- https://github.com/cultuurnet/udb3-search-service/blob/main/src/ElasticSearch/Operations/json/mapping_organizer.json

projects/uitdatabank/docs/search-api/advanced-queries.md
bertramakers: Or, you could expand the example to also look for the Paspartoe label of course

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,997 @@
+# Advanced queries
+
+With advanced queries you can make more complex queries, using boolean operators and a specific set of parameters. 
+
+The syntax is based on the Lucene query syntax. More info about the syntax can be found in the [ElasticSearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#query-string-syntax).
+
+## Parameters
```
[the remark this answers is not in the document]
corneelwille: This should be fixed in a new commit, for all fields

projects/uitdatabank/docs/search-api/advanced-queries.md
corneelwille: Not in the JSON of the event or in Search API. However, if you create an event / place / organizer / image or mediaObject / video through the API the response will look like this: 

```
{
    "id": "affcc3d4-95bb-46cf-8095-cde4d755f351",
    "placeId": "affcc3d4-95bb-46cf-8095-cde4d755f351",
    "url": "https://io-acc.uitdatabank.be/place/affcc3d4-95bb-46cf-8095-cde4d755f351",
    "commandId": "00000000-0000-0000-0000-000000000000"
}
```

But I see this is not clear in the documentation, so it needs some rework

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -380,7 +381,11 @@ GET /organizers/?q=imagesCount:[5 TO 10]
 
 ### labels
 
+<<<<<<< HEAD
+Search for documents that have a specific label with the `labels` field.
+=======
 Search for documents that have a specific label (or hidden label) with the `labels` parameter.
+>>>>>>> b52c705c8e68707cb8a713bc93287b86f9a2e92a
```
bertramakers, line 391: There are some artifacts from a merge conflict left over here. I think the sentence should be:

```
Search for documents that have a specific label (or hidden label) with the `labels` field.
```

(So a mix of the two)

projects/uitdatabank/docs/search-api/advanced-queries.md
bertramakers: I understand it now based on your comment but I think it's still unclear in the documentation. I think it would be clear if we reword:

> looking for the value for the `id` property in the JSON response of the UiTdatabank entry API

to:

> looking for the value for the `id` property in the JSON response of the UiTdatabank entry API to a `POST /events` request (to create a new event)

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,991 @@
...
+
+With the `_exists_` field you can search for the documents that have a specific property in their JSON body.
+
+**Applicable on endpoints**
+
+`/events` `/places` `/offers` `/organizers`
+
+**Possible values**
+
+Any field name, as documented on this page.
```
bertramakers, line 311:
```suggestion
Any field name documented on this page.
```

Very small nit, but says the same in a shorter sentence without a comma

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1067 @@
...
+
+The following example looks for matches in any language:
+
+```
+GET /places/?q=address.\*.addressCountry:BE
+```
+
+### address.{language}.addressLocality
+
+With the `addressLocality` field you can limit your results to one or more municipalities.
```
bertramakers, line 39: (Copied from older comment on previous filename)

Should we suggest to use the `postalCode` parameter instead when possible? Since the municipality names are free text in the API, so they may contain typo's or be incorrect.

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1067 @@
...
+### calendarType
+
+Every event and place in UiTdatabank has one of the following four calendarTypes:
+
+* `Single`: the event occurs on a single date, indicated by a single startDate and endDate.
+* `Multiple`: the event occurs on multiple dates, and has multiple subEvent entries with each a different startDate and endDate.
+* `Periodic`: the event or place runs for a specific period as indicated by its startDate and endDate, and can optionally have openingHours.
+* `Permanent`: the event or place is permanent and has no startDate or endDate, but it can optionally have openingHours.
+
+With the `calendarType` field you can look for results that match a certain calendarType.
```
bertramakers, line 198: (Copied from older comment on previous filename)

I think it's best to not capitalize the types here, as they are always lowercase in the APIs

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1067 @@
...
+TO DO: link to booking availability guide when that guide is finished.
+
+### calendarType
+
+Every event and place in UiTdatabank has one of the following four calendarTypes:
+
+* `Single`: the event occurs on a single date, indicated by a single startDate and endDate.
+* `Multiple`: the event occurs on multiple dates, and has multiple subEvent entries with each a different startDate and endDate.
+* `Periodic`: the event or place runs for a specific period as indicated by its startDate and endDate, and can optionally have openingHours.
+* `Permanent`: the event or place is permanent and has no startDate or endDate, but it can optionally have openingHours.
```
bertramakers, line 196: (Copied from older comment on previous filename)

I think it's best to not capitalize the types here, as they are always lowercase in the APIs

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1067 @@
...
+
+`/events` `/places` `/offers` `/organizers`
+
+**Possible values**
+
+Any ISO-8601 datetime, e.g. `2023-04-01T12:08:01+01:00`
+
+<!-- theme: info -->
+
+> Note that the `+` sign must be encoded for URLs (as `%2B`). Otherwise it will be interpreted as whitespace and the given date time will be considered invalid.
```
bertramakers, line 254: (Copied from older comment on previous filename)

I think technically everything inside the `q` must be URL-encoded.

For example when I go to [https://search.uitdatabank.be/offers/?q=labels:"UiTPAS Leuven"](https://search.uitdatabank.be/offers/?q=labels:%22UiTPAS%20Leuven%22) I can see that Firefox changes it to https://search.uitdatabank.be/offers/?q=labels:%22UiTPAS%20Leuven%22 "under the hood" (but in the URL bar it will not change as it will show it as "human-readable" to users).

So maybe something to mention at the start of the page (or in the escaping guide and call that for example "escaping & encoding"), and then mention here something like:

> Note that because everything inside `q` must be URL-encoded, the `+` sign should be encoded for as `%2B`. Otherwise it will be interpreted as whitespace and the given date time will be considered invalid.

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1067 @@
...
+
+Use the `creator` field to search for documents created by a specific creator.
+
+**Applicable on endpoints**
+
+`/events` `/places` `/offers` `/organizers`
+
+**Possible values**
+
+Any uuid of a creator.
```
bertramakers, line 274: (Copied from older comment on previous filename)

It is only a uuid in the case of UiTID v1 creators. In the case of UiTID v2 users (that did not exist in UiTID v1), it can be a lot of things. For example:

- `google-oauth2|117412536054327843587` for Google logins
- `facebook|10159360927408483` for Facebook logins
- `auth0|5f03150966abde00196567e5` for email + password logins

And for m2m tokens it will be something like `Nnx3vGNuVVX2KJdtzHhwnx15WlTqGCZ9@clients`

(I've used fake ids here since this is a public repo)

I wouldn't try to explain all of these different formats as we cannot control them. Instead we can explain that they can look at the `creator` field of events/places, or the `id` field when they access the `/user` endpoint with  a user token. For client tokens there is currently no good way to find the value that they creator would have for a given token, but we could add one like a `/client` endpoint

Also if we specify that this is specifically a UUID, some integrators may depend on that format in their code (for example I've seen an integrator complain that some events did not have a valid UUID as id because they typed their database column as a hexadecimal field specifically, while some events from UDB2 do not have a valid UUID and thus would not be accepted by that database field). It's better to just treat this as an opaque string that can be anything, and then it doesn't matter to integrators if it's sometimes a UUID, an email address, or one of those auth0 ids.

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1067 @@
...
+
+**Example**
+
+Retrieve all events that took place in the year 2020:
+
+```
+GET /events/?q=dateRange:[2020-01-01T00\:00\:00%2B01\:00 TO 2021-12-31T23\:59\:59%2B01\:00]&availableTo=*&availableFrom=*
+```
+
+### \_exists\_
```
bertramakers, line 310: (Copied from older comment on previous filename)

Maybe put this at the top or bottom, as alphabetical sorting usually does for words that start with special characters

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1067 @@
...
+
+With the `_exists_` field you can search for the documents that have a specific property in their JSON body.
```
bertramakers, line 312: (Copied from older comment on previous filename)

Technically it returns documents that have the given field name in their index. Which is a bit different, because we do not index all JSON fields, or we index them with a slightly different name.

For example all events have a `sameAs` property with a link to UiTinVlaanderen, but we don't index that so `_exists_:sameAs` will return 0 results.

And the reverse is also true: `_exists_:address` will return places as well as events, while events never have an `address` property in their JSON. Because we index an event's location's address as `address`, so you can search on `address` in both events and places at the same time

If it helps to visualize it, you can find the properties that we index for events here: https://github.com/cultuurnet/udb3-search-service/blob/main/src/ElasticSearch/Operations/json/mapping_event.json

So what we do is we take the UDB JSON, then transform it to a new JSON document that has all or some of the properties listed in this mapping file. The mapping file is then used to tell Elasticsearch how to analyze specific fields. (Not how to transform the UDB JSON to the JSON documents in Elasticsearch!)

Basically what `_exists_` then does is check if a JSON document in Elasticsearch has one of these properties from the mapping file.

And it's also these field names that you can filter on in the advanced queries, if their analyzer supports it. (For example you cannot filter on `geo` or `geo_point` in the advanced queries because the `geo_shape` and `geo_point` field types are not supported in "query string queries" (= Elasticsearch name for what we call "advanced queries")

Here are the ones for places and organizers:

- https://github.com/cultuurnet/udb3-search-service/blob/main/src/ElasticSearch/Operations/json/mapping_place.json
- https://github.com/cultuurnet/udb3-search-service/blob/main/src/ElasticSearch/Operations/json/mapping_organizer.json

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -0,0 +1,1067 @@
...
+
+You can use wildcards in field names to search across multiple translated versions. When doing so, the wildcard must be escaped using a backslash, like in the example below:
+
+```
+GET /events/q=address.\*.postalCode:9000
+```
+
+**Using date time filters**
+
+A colon (`:`) must always be escaped using a backlash (`\`), e.g. in a date time filter:
```
bertramakers, line 1056: As mentioned on Slack I don't think this is 100% correct. For example this works (and is exactly the same as the example below but without the escaping):

```
https://search.uitdatabank.be/events/?disableDefaultFilters=true&q=dateRange:[2022-01-01T00:00:00%2B01:00%20TO%202022-12-31T23:59:59%2B01:00]
```

projects/uitdatabank/docs/search-api/advanced-queries.md
corneelwille: I will add this in the escaping (& encoding) guide

projects/uitdatabank/docs/search-api/advanced-queries.md
corneelwille: I replaced `uuid` with `id` or `identifier` (according to the context / sentence)

projects/uitdatabank/docs/search-api/advanced-queries.md
corneelwille: Not sure how to solve this. I think adding a list of all the properties in the mapping file would be "overdocumenting" it. Especially since the `_exists_` parameter is not used that much afaik (since it is not documented in the current documentation).

bertramakers: I think we could just mention that you could check any field name mentioned on this page?

corneelwille: That is already documented: https://github.com/cultuurnet/apidocs/pull/65/files#diff-556cbdc659d1c83f5dd74ba5bb8fc0cf1130101335165c8547d491f0fc7a4c6dR962-R964

```
**Possible values**

Any field name documented on this page.
```

projects/uitdatabank/docs/search-api/advanced-queries.md
bertramakers: Clarified in 747b07e5031d75a266c8192ffb3dfe6d1fdd92fd I see 👍

projects/uitdatabank/docs/search-api/advanced-queries.md
bertramakers: Fixed in fd60622e4f3630aa90380419d0ae92f5486c6611 I see 👍

projects/uitdatabank/docs/search-api/advanced-queries.md
bertramakers: Fixed in b6929769ef513c5c3f2afedcf4732019aec84a73 I see 👍

projects/uitdatabank/docs/search-api/advanced-queries.md
bertramakers: Ah yes you're right I overlooked that. Still, the "search for the documents that have a specific property in their JSON body" at the start can make some integrators think that they can search for any property in the JSON of events/places/organizers while it's only the fields documented on this page? So we could change that to for example `search for the documents that have a value for one of the fields documented on this page`?

bertramakers: Oh you changed it in 67947e2bd87e93364d0ebca13abc41dde117624f 👍  (I'm reviewing commit-per-commit in this PR since it's so big)

projects/uitdatabank/docs/search-api/advanced-queries.md
```diff
@@ -105,7 +105,7 @@ A street name and (!) a street number.
 Retrieve all organizers that are located on Henegouwenkaai 41-43:
 
 ```
-GET /organizers/q=address.nl.streetAddress:"Henegouwenkaai 41-43"
+GET /organizers/?q=address.nl.streetAddress:"Henegouwenkaai 41-43"
```
bertramakers, line 108: Very well spotted! 🕵️‍♂️

bertramakers (approved):
Beautiful! ✨  Let's get this merged and published 🚢

PR #72, opened 2022-09-28
Uitpas/ups 3504 extra

erwin1:
### Changed

- fixed array type for city and postalCode
- documented 404
- added allowsCardlessRegistration boolean

---

Ticket: https://jira.uitdatabank.be/browse/UPS-3504

### DIFF
--- a/projects/uitpas/reference/uitpas.json
+++ b/projects/uitpas/reference/uitpas.json
@@ -4354,7 +4354,14 @@
             "$ref": "#/components/responses/Forbidden"
           },
           "404": {
-            "description": "Not Found. Error type:\n\nhttps://api.publiq.be/probs/url/not-found"
+            "description": "Not Found. Error type:\n\nhttps://api.publiq.be/probs/url/not-found",
+            "content": {
+              "application/problem+json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            }
           }
         },
         "operationId": "get-card-systems",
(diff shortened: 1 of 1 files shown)


projects/uitpas/reference/uitpas.json
```diff
@@ -4378,22 +4378,32 @@
             },
             "description": "Maximum amount of results to return. Can be used in combination with `start` for pagination.",
             "name": "limit"
-          },
+          },
```
bertramakers: Nit: These trailing whitespaces can be removed

bertramakers: Ah resolved in the next commit

projects/uitpas/reference/uitpas.json
```diff
@@ -4352,6 +4352,9 @@
           },
           "403": {
             "$ref": "#/components/responses/Forbidden"
+          },
+          "404": {
+            "description": "Not Found. Error type:\n\nhttps://api.publiq.be/probs/url/not-found"
```
bertramakers: In which case would this endpoint return a 404? If there are no results found I'd expect it to return 200 but with an empty list.

bertramakers (approved):
Not 100% sure when `GET /card-systems` would return a 404, but other than that it looks good!

projects/uitpas/reference/uitpas.json
erwin1: this can be the case when postalCode or city does not validate

PR #73, opened 2022-10-06
Document error type for when an event has UiTPAS ticket sales

bertramakers:
### Added

- New error type in UiTdatabank docs for when an event has UiTPAS ticket sales

---

Ticket: https://jira.uitdatabank.be/browse/III-4758


corneelwille (approved):
Super clear 👌

JonasVHG (approved):

PR #78, opened 2022-10-21
API-100 Fix `stoplight-id` frontmatter being altered by `yarn docs:lint:fix`

bertramakers:
### Fixed

- Running `yarn docs:lint:fix` no longer reformats the frontmatter added by Stoplight Studio (since a recent update) in a way that turns it into content.

---

Ticket: https://jira.uitdatabank.be/browse/API-100

More details: Since a recent update Stoplight Studio automatically adds some Markdown frontmatter to every new article that you create (when using "Create new article"):

```
---
stoplight-id: d6a8ba3f3c186
---
```

Running `yarn docs:lint` reports no errors on this, but when you run `yarn docs:lint:fix` it changed the frontmatter to:

```
***

# stoplight-id: d6a8ba3f3c186
```

This made the `stoplight-id` a H1 header, which we obviously don't want.

The easy workaround was to just remove the frontmatter, since the preferred way to have "pretty page URLs" is not to modify the `stoplight-id`, but to use a `slug` in the `toc.json`. (See README)

However it's annoying for contributors who don't know about this bug and/or don't know that they can just remove this frontmatter. By adding the `remark-frontmatter` plugin, we can make remark ignore it when running the fixer and we avoid the frontmatter turning into a H1 header.

### DIFF
--- a/projects/api-design-guidelines/docs/introduction.md
+++ b/projects/api-design-guidelines/docs/introduction.md
@@ -1,3 +1,7 @@
+---
+stoplight-id: d6a8ba3f3c186
+---
+
 # Introduction
 
 Welcome to publiq's API design and documentation guidelines! 👋
--- a/.remarkrc
+++ b/.remarkrc
@@ -1,5 +1,6 @@
 {
   "plugins": {
+    "remark-frontmatter": null,
     "remark-preset-lint-recommended": null,
     "remark-lint-no-dead-urls": {
         "skipUrlPatterns": [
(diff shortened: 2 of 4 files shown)


projects/api-design-guidelines/docs/introduction.md
```diff
@@ -1,3 +1,7 @@
+---
+stoplight-id: d6a8ba3f3c186
+---
```
bertramakers, line 3: This already was the id of this page on docs.publiq.be, I just added it as a proof-of-concept to show that `yarn docs:lint:fix` does not alter it.

PR #79, opened 2022-10-24
API-65 Add property for end-user error messages

bertramakers:
### Added

- Added `endUserMessage` property on error response schema

### Changed

- Clarified some existing properties

---

Ticket: https://jira.uitdatabank.be/browse/API-65

### DIFF
--- a/projects/errors/docs/format.md
+++ b/projects/errors/docs/format.md
@@ -48,9 +48,9 @@ Aside from the standardized properties above, the RFC7807 spec also allows for c
 
 #### endUserMessage
 
-Error responses may include an `endUserMessage` property with a message that explains the problem to an end-user, localized in one or more languages (depending on the target audience of the API). These are usually included in domain errors, for example when an UiTPAS passholder has an expired social tariff.
+Error responses may include an `endUserMessage` property with a message that explains the problem to an end-user, localized in one or more languages (depending on the target audience of the API). It is usually included in domain errors, for example when an UiTPAS passholder has an expired social tariff.
 
-When an `endUserMessage` is included in the error response, it is strongly encouraged to always show these to end-users.
+When an `endUserMessage` is included in the error response, it is strongly encouraged to show it to the end-user.
 
 An example response with an `endUserMessage` looks like:
 


bertramakers:
@erwin1 @LucWollants I will update the markdown files to also further document this property, but can you first review it and let me know if you can work with this in the case of UiTPAS @erwin1 ? So I don't spend time on documenting a property that needs to be changed after all.

erwin1:
I gave it a try: no problem to use this 👍

LucWollants (approved):
Very clear property names and messages.

bertramakers:
Okay thanks! I will now add the extra documentation for it and then ask for another review :)

bertramakers:
@LucWollants @erwin1 I've also updated the documentation in the "Errors" project and the internal API design guidelines to explain this new property and when to include it (as well as clarifying when to use `schemaErrors`)

projects/errors/docs/format.md
```diff
@@ -0,0 +1,100 @@
+# Response format & examples
+
+Error responses from publiq APIs follow the [RFC7807](https://datatracker.ietf.org/doc/html/rfc7807) standard, also known as "Problem Details for HTTP APIs".
```
erwin1, line 3: Is it clear for client developers what "publiq APIs" are?

Maybe we need an extra section somewhere else (maybe on the doc home page) to clarify this. Or maybe we can clarify it here.

projects/errors/models/Error.json
```diff
@@ -9,19 +9,42 @@
...
       "description": "The HTTP status code."
     },
     "detail": {
       "type": "string",
-      "description": "A human-readable explanation specific to this occurrence of the problem. "
+      "description": "A human-readable explanation specific to this occurrence of the problem (for developers). "
+    },
+    "endUserMessage": {
+      "type": "object",
+      "description": "A human-readable explanation of the problem, specifically for end-users, in one or more languages. Usually set for domain errors, but not for errors caused by a technical issue in the integration (for example invalid JSON syntax in a request body). An `nl` value is always provided, other languages may be provided depending on the API and its intended audience.",
```
erwin1, line 24: I would repeat the phrase about recommendation to show to end-users here as well:
```suggestion
      "description": "A human-readable explanation of the problem, specifically for end-users, in one or more languages. Typically available for domain errors, but not for errors caused by a technical issue in the integration (for example invalid JSON syntax in a request body). An `nl` value is always provided, other languages may be provided depending on the API and its intended audience. When this property is included, it is strongly encouraged to show this to the end-user.",
```

projects/errors/docs/format.md
```diff
@@ -0,0 +1,100 @@
...
+}
+```
+
+### Extensions
+
+Aside from the standardized properties above, the RFC7807 spec also allows for custom properties. The following extra properties are sometimes used on publiq APIs to provide more information in specific situations.
+
+#### endUserMessage
+
+Error responses may include an `endUserMessage` property with a message that explains the problem to an end-user, localized in one or more languages (depending on the target audience of the API). These are usually included in domain errors, for example when an UiTPAS passholder has an expired social tariff.
```
erwin1, line 51: single end user message?

```suggestion
Error responses may include an `endUserMessage` property with a message that explains the problem to an end-user, localized in one or more languages (depending on the target audience of the API). It is usually included in domain errors, for example when an UiTPAS passholder has an expired social tariff.
```

projects/errors/docs/format.md
```diff
@@ -0,0 +1,100 @@
...
+
+When an `endUserMessage` is included in the error response, it is strongly encouraged to always show these to end-users.
```
erwin1, line 53:
```suggestion
When an `endUserMessage` is included in the error response, it is strongly encouraged to show it to the end-user.
```

projects/errors/docs/format.md
bertramakers: This was just moved from existing documentation to a new page, and we use this in other places as well. I think it should become clear once we have "publiq platform", as well as from the URL (docs.publiq.be) and the logo in the top left (which mentions "publiq").

bertramakers: But the home page will be expanded before we release publiq platform / these docs (to completely replace documentatie.uitdatabank.be) - so that can also further clarify it yes

PR #86, opened 2022-10-27
Uitdatabank - Add schema to validate event and place image update endpoint

steftrenson:
### Added
- schema to validate event and place image update endpoint

---

Ticket: https://jira.uitdatabank.be/browse/III-4258

### DIFF
--- a/projects/uitdatabank/models/event-image-put.json
+++ b/projects/uitdatabank/models/event-image-put.json
@@ -1,0 +1,26 @@
+{
+  "title": "event.image.put",
+  "type": "object",
+  "description": "Describes the request body to update an image at an [event](./event.json).",
+  "examples": [],
+  "properties": {
+    "description": {
+      "type": "string",
+      "example": "Picture of the last publiq event",
+      "minLength": 1,
+      "pattern": "\\S",
+      "description": "Description of the image object."
+    },
+    "copyrightHolder": {
+      "type": "string",
+      "example": "publiq",
+      "minLength": 3,
+      "pattern": "\\S",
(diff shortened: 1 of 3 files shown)


projects/uitdatabank/models/event-image-put.json
```diff
@@ -0,0 +1,26 @@
+{
+  "title": "event.image.put",
+  "type": "object",
+  "description": "Describes the request body to update an image at an [event](./event.json).",
+  "examples": [],
+  "properties": {
+    "description": {
+      "type": "string",
+      "example": "Picture of the last publiq event",
```
bertramakers, line 9: It is preferred to add examples to the `examples` property (set on the schema as well but an empty list right now) instead of per property. See also https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0 ("Use examples not example")

projects/uitdatabank/models/event-image-put.json
```diff
@@ -0,0 +1,26 @@
+{
+  "title": "event.image.put",
+  "type": "object",
+  "description": "Describes the request body to update an image at an [event](./event.json).",
+  "examples": [],
+  "properties": {
+    "description": {
```
bertramakers, line 7: You could use a `$ref` to existing schemas for both these properties I think? 

- https://github.com/cultuurnet/apidocs/blob/main/projects/uitdatabank/models/common-description-localized.json
- https://github.com/cultuurnet/apidocs/blob/main/projects/uitdatabank/models/common-copyrightHolder.json

PR #88, opened 2022-10-28
Uitdatabank/requirements before going live

corneelwille:
### Changed
Minor tweaks:
* Fix typo
* Remove sentence in step 3 that is not in line with the info in step 1


bertramakers (approved):

bertramakers:
FYI your commit had a failing linter check because you were missing the commit with the fix for the link to the taxonomy terms guide. I think because you re-created the branch from a local copy which was not 100% up-to-date. I've merged main into it and that fixed it. Just sharing so you know how you can fix it if you run into it again in the future :)

PR #91, opened 2022-11-03
API-103 Linter optimization

bertramakers:
### Fixed

- `yarn docs:lint` now reports most changes that `yarn docs:lint:fix` would make. For example, in the past `yarn docs:lint:fix` would change the marker of unordered list items from `-` to `*`, but `yarn docs:lint` would not report this. So the automatic checks for a PR could be green, but after merging it and running `yarn docs:lint:fix` in a different branch/PR it would suddenly make changes to files added/edited in the previous PR (sometimes even outside the project of the 2nd PR).

### Changed

- Lists must no longer be indented with 4 characters, but only 2. (So as `* example`)
- Code blocks must now always be fenced with backticks (` ``` `) instead of either being fenced (with multiple possible patterns) or being indented with 4 spaces
- Unordered list markers MUST now use `*`
- Configured formatting of other structures in case we'd ever use them, but of no consequence to existing files

(The rules above, and many others, can automatically be fixed with `yarn docs:lint:fix`)

---

Ticket: https://jira.uitdatabank.be/browse/API-103


bertramakers:
Because the changes in this PR are only automatic linting fixes (aside from the extra linter config to fix as many discrepancies between `yarn docs:lint` and `yarn docs:lint:fix` as possible), I'm going to merge this without an explicit approval to reduce merge conflicts with other PRs. If anyone has any feedback it can always be processed in a follow-up PR.

PR #93, opened 2022-11-04
API-106 Allow GFM syntax in Markdown files

bertramakers:
### Added

- Added support for [GFM](https://github.github.com/gfm/) syntax in Markdown files

---

Ticket: https://jira.uitdatabank.be/browse/API-106


bertramakers:
As with #91 I'm going to merge this without explicit approvals, as otherwise this could stay open for some time and it would lead to merge conflicts. All changes are just automatic formatting by the linter now that GFM is supported, and I've double checked them on docs.publiq.be

PR #94, opened 2022-11-07
Uitdatabank - Add schema to validate event and place update contactPoint endpoint

steftrenson:
### Added
- Add schema to validate event and place update contactPoint endpoint

---

Ticket: https://jira.uitdatabank.be/browse/III-4263

### DIFF
--- a/projects/uitdatabank/models/event-contactPoint-put.json
+++ b/projects/uitdatabank/models/event-contactPoint-put.json
@@ -8,9 +8,7 @@
       "minItems": 0,
       "items": {
         "type": "string",
-        "description": "Phone number for contact purposes.",
-        "minLength": 1,
-        "pattern": "\\S"
+        "description": "Phone number for contact purposes."
       }
     },
     "email": {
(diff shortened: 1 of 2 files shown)


bertramakers (approved):

projects/uitdatabank/models/event-contactPoint-put.json
```diff
@@ -1,4 +1,53 @@
...
+  "properties": {
+    "phone": {
+      "type": "array",
+      "description": "List of phone numbers for contact purposes (can be empty)",
+      "minItems": 0,
+      "items": {
+        "type": "string",
+        "description": "Phone number for contact purposes.",
+        "minLength": 1,
+        "pattern": "\\S"
```
bertramakers, line 13: Are these min-length and "at least one non-whitespace character required" checks also in the current code for this endpoint? (Or do you get a 500 if you pass an empty string or a string with only whitespace?) If not I'd leave them out.

bertramakers (approved):

PR #97, opened 2022-11-09
correct error location

erwin1:
### Fixed

- Fixed location to Error.json


bertramakers (approved):
Good find 👍 I double checked and I don't see any other files that contains `cultuurnet/stoplight-docs` so hopefully this is the last of the references to those older repositories

PR #108, opened 2022-11-17
Add language property to field name

corneelwille:
### Added
One more missing language property in a field name

### DIFF
--- a/projects/uitdatabank/docs/errors.md
+++ b/projects/uitdatabank/docs/errors.md
@@ -98,8 +98,8 @@ The action you are trying to perform on an event cannot be done because the even
 
 ## image-must-be-linked-to-resource
 
-*   **Complete type:** `https://api.publiq.be/probs/uitdatabank/image-must-be-linked-to-resource`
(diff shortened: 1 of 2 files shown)


projects/uitdatabank/docs/errors.md
```diff
@@ -98,8 +98,8 @@ The action you are trying to perform on an event cannot be done because the even
 
 ## image-must-be-linked-to-resource
 
-*   **Complete type:** `https://api.publiq.be/probs/uitdatabank/image-must-be-linked-to-resource`
-*   **Title**: `Image must be linked to resource`
-*   **Status**: `400`
+* **Complete type:** `https://api.publiq.be/probs/uitdatabank/image-must-be-linked-to-resource`
+* **Title**: `Image must be linked to resource`
+* **Status**: `400`
```
bertramakers, line 103: (Note: This was an issue on `main` because an open PR that was checked with the old linting rules was merged, it's also fixed in #110 but it's okay to leave it in here as well)

bertramakers (approved):

PR #111, opened 2022-11-18
Create getting started page

corneelwille:
### Added
High level overview of the process of configuring and embedding a widget.

### DIFF
--- a/projects/widgets/docs/getting-started.md
+++ b/projects/widgets/docs/getting-started.md
@@ -6,7 +6,7 @@ Before you can start building your widget you must make a widget project on our 
 
 The created widget project will use event data from our test environment. No payment is required until you decide you want to use real event data.
 
-https://youtu.be/P5RHLT7Q37E
+<https://youtu.be/P5RHLT7Q37E>
 
 <!-- theme: success -->
 


projects/widgets/docs/getting-started.md
```diff
@@ -1,3 +1,48 @@
-# getting started
+# Getting started
 
-...
+Once you have created your widget project on [Projectaanvraag](https://projectaanvraag.uitdatabank.be), you're all set to start building your widget.
+
+Below we explain the different steps of the configuration process in a nutshell.
```
bertramakers, line 5: This seems more like the "configuration" guide than the "getting started" guide? Getting started should probably explain that you can create your project on projectaanvraag, and "configuration" can cover this "configuration process" that you mention here?

I'll review the rest later once this is cleared up to avoid potentially needless comments.

projects/widgets/docs/getting-started.md
```diff
@@ -1,3 +1,66 @@
...
+# Getting started
 
-...
+## 1. Registering a widget project
+
+Before you can start building your widget you must make a widget project on our [self-service platform Projectaanvraag](https://projectaanvraag.uitdatabank.be). The registration process is very simple and only takes a few minutes.
+
+The created widget project will use event data from our test environment. No payment is required until you decide you want to use real event data.
+
+<https://www.youtube.com/watch?v=X0uplBw98gg>
```
corneelwille, line 9: FYI: the screen capture could be better (e.g. it stops quite abruptly) but since we are building a new Projectaanvraag I didn't invest time in getting the screen capture better

projects/widgets/docs/getting-started.md
```diff
@@ -1,3 +1,66 @@
...
+
+### Styling your widget 🎨
+
+Once embedded in a website, the widget will automatically inherit the CSS styles of the guest website.
+
+In addition, you can also define your own CSS rules in the widget builder to finetune the widget to the design of your website. If CSS is not your cup of coffee, you can also use one of the predefined themes we provide in the widget builder.
+
+👉 [Learn more about widget styling](./styling.md)
+
+## 3. Activating your widget project to use real event data
```
corneelwille, line 60: When publiq platform is released we can add a screen capture of the activation too

projects/widgets/docs/getting-started.md
```diff
@@ -1,3 +1,66 @@
...
+
+> Once you have created your widget project, you're all set to start building your widget!
+
+## 2. Building your widget
+
+### Choosing a template 🗂️
+
+To help you get started quickly, we offer 4 different templates:
+
+1. **Leeg**: You start from a blanc layout that you then expand yourself by adding rows and assigning blocks (e.g. search box or search results) to these rows.
```
bertramakers, line 21:
```suggestion
1. **Leeg**: You start from a blank layout that you then expand yourself by adding rows and assigning blocks (e.g. search box or search results) to these rows.
```

At least I think it should be `blank`. See https://en.wiktionary.org/wiki/blanc#English

If you have a definition that explains that it should be `blanc` in some cases (in English), do share!

projects/widgets/docs/getting-started.md
```diff
@@ -1,3 +1,66 @@
...
+* which events you want to display in your agenda
+* which info you want to show about the events
+
+The way of configuring is very simple and is mainly done by ticking and unchecking checkboxes.
+
+👉 [Learn more about the  configuration settings](./configuration.md)
+
+### Publishing and embedding your widget 💾
+
+Once you're satisfied with the configuration you've done in step 2 you can publish your widget. The publish action will save all configuration changes you've made on the `<script>` tag of your widget.
```
bertramakers, line 46: The "save all configuration changes you've made on the `<script>` tag of your widget" might make it seem like all configuration is stored inside the `<script>` tag (which you copy & paste on your website), and you will need to update the `<script>` tag on your website when you update the configuration? At least that's how I could see it being interpreted but it's subjective of course.

bertramakers: You could of course just mention somewhere (for example after the next sentence below) that you only need to embed the script tag once, and that you don't need to change it after every configuration update

projects/widgets/docs/getting-started.md
```diff
@@ -1,3 +1,66 @@
...
+
+Now you can copy the unique `<script>` tag to embed it on the page of your website where you want to show the agenda.
+
+👉 [Learn more about publishing & embedding](./publish-and-embed-on-your-website.md)
+
+### Styling your widget 🎨
+
+Once embedded in a website, the widget will automatically inherit the CSS styles of the guest website.
+
+In addition, you can also define your own CSS rules in the widget builder to finetune the widget to the design of your website. If CSS is not your cup of coffee, you can also use one of the predefined themes we provide in the widget builder.
```
bertramakers, line 56:
```suggestion
In addition, you can also define your own CSS rules in the widget builder to finetune the widget to the design of your website. If CSS is not your cup of tea, you can also use one of the predefined themes we provide in the widget builder.
```

Might seem pedantic but "your cup of tea" is an idiom and "your cup of coffee" is not: https://www.quora.com/Why-do-you-say-it-is-not-my-cup-of-tea-but-never-it-is-not-my-cup-of-coffee-to-denote-something-you-dont-like-What-is-the-origin-of-the-idiom

But now I'm nitpicking of course so if you want to leave it like this I don't see any harm.

projects/widgets/docs/getting-started.md
bertramakers: Clear now!

projects/widgets/docs/getting-started.md
corneelwille: Ah, good to know 👌

bertramakers (approved):

PR #118, opened 2022-11-22
Uitdatabank/entry api guides school events

corneelwille:
### Added
- Documentation page on how to create a school event with the UiTdatabank Entry API

### Some context & remarks for the reviewer
For this pull request I used the current / existing documentation available on https://documentatie.uitdatabank.be/content/content-management/latest/school-offers.html as a starting point.

I did not change any substantive. However I tried to:
* improve the definitions of the different types of school events
* provide an alternative for the complex table structure for the different labels 

I'm not familiar with all the ins and outs of the flow for school events, so I think an extra functional review by Eva (Product manager Cultuurkuur) is desirable. Especially since I have the feeling that some of the cultuurkuur labels are on the verge of getting outdated already, with the "onderwijsvernieuwing" that is going on (see https://www.kwalificatiesencurriculum.be/sites/default/files/2022-06/Poster%20matrix_0.pdf): 
* terms like "ASO" and "BSO" are getting replaced by different "finaliteiten"? 
* also the "domeinen" could possibly have an impact on the `subject labels` that are being used by Cultuurkuur.


projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
+# School events
+
+School events are events that are specifically organized and intented for:
+
+* students (e.g. a school performance)
+* teachers (e.g. an after-school training)
+
+School events are published on [Cultuurkuur](https://www.cultuurkuur.be), but not on UiTinVlaanderen.
+
+> If you’re planning to create school events in UiTdatabank with the Entry API, please contact <content.cultuurkuur@publiq.be> in advance so that we can guide you through this process.
```
bertramakers, line 10: What does this guidance include that is not included in this documentation? Or is it more that we want to have an idea who is going to implement this and they need to contact us so we know that they are going to implement this?

Because it seems like this documentation covers everything, so it's not unlikely that someone is going to implement it like it documented here without contacting us. So if you want to be sure they contact us for whatever reason, I think we need to clarify here that we will either provide extra info that is not provided here and that they should have (although I'm not sure why we cannot document it here then), OR they need to contact us to notify us that they are going to implement this so we can do some kind of review etc.

Alternatively if you meant that they can optionally contact us about any questions they might have related to school events, I'd rephrase it to clarify that.

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
+# School events
+
+School events are events that are specifically organized and intented for:
```
bertramakers, line 3:
```suggestion
School events are events that are specifically organized and intended for:
```

Small typo

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+
+We distinguish 3 different types of school events:
+
+1. **School performances** are events of which both the date and the location is known in advance.
+2. **Guided tours** are events that have no specific date, but they do have a location.
+3. **Bookable events** are events that have no specific date and location (or the date and location is not known in advance). Both the date and the location are determined in mutual agreement between the organizer or artist and the consumer (school).
+
+### School performances
+
+School performances are events of which both the date and the location is known in advance. For example, a theater performance aimed at a toddler of 3-4 years old in hetpaleis on 14/05/2023, 14:30 - 16:00 PM.
```
bertramakers, line 22: Maybe it's me but I didn't know what "hetpaleis" was and considered it to be a typo. Maybe make it a link to their website, or put it in quotes (`"hetpaleis"`), or use another example (preferably one that is clearly a name, e.g. starts with a capital letter)?

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+
+* ✅ date is known in advance
+* ✅ location is known in advance
+
+Since school performances both have a date and a location, you can create them in a very similar way to regular events:
+
+* you can use a calendarType of your preference (`single`, `multiple` or `periodic`)
+* for the location you must use the URL of an existing place as `location.@id` in the `POST /events` request of the event
```
bertramakers, line 30: Maybe link to the guide about re-using places here?

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+
+For example, as a school you can book Stijn Meuris for a school performance at your school or local cultural centre, on a date that suits both the artist and the school.
+
+* ❌ date is not known in advance
+* ❌ location is not known in advance
+
+In order to create a bookable school event you must use
+
+* calendarType `permanent`
+* the URL of the dummy place as `location.@id` in the `POST /events` request of the bookable events
```
bertramakers, line 56: This is called a "location in consultation with the school" in the terminology page. https://docs.publiq.be/docs/uitdatabank/76b7bd22f7f51-terminology#location-in-consultation-with-the-school

We can definitely rename that (I just based it on the "locatie in overleg met school"), but I would not use the name "dummy location" for this because we also use that for locations that have no `@id`: https://docs.publiq.be/docs/uitdatabank/76b7bd22f7f51-terminology#dummy-location

(Other occurrences of "dummy location" in this guide about school events should also be renamed if you agree)

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+**URL of the dummy place for bookable events**:
+
+* Test environment: `https://io-test.uitdatabank.be/place/3b92c85b-a923-4895-85f5-ed056dae11e2`
+* Production environment: `https://io.uitdatabank.be/place/c3f9278e-228b-4199-8f9a-b9716a17e58f`
+
+## Creating school events
+
+For the creation of school events several extra requirements apply:
+
+1. The `audienceType` must be set to `education`
```
bertramakers, line 67: I think this is technically not required when using a "location in consultation with the school" `@id` (aka the "dummy location" mentioned right now) because UDB3 then automatically sets the audienceType to "education" and also prevents it from being changed, but on the other hand it does not harm either and for school performances and guided tours it needs to be set explicitly of course. So you could mention that it's not required in that one case of bookable events, but that might also make it more confusing. Just mentioning it so you can decide with all the possible info.

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+2. The event must have an `organizer` that has the `Cultuurkuur` label
+3. Specific education related `labels` are mandatory
+4. In case of a [guided tour](#guided-tours) or a [bookable event](#bookable-events): the `calendarType` must be set to `permanent`
+5. In case of a [bookable event](#bookable-events): the `dummy place for bookable events` must be used for the location
+
+### audienceType
+
+For school events you must include an extra property `audienceType` and set the value for the property to `education`.
+
+```js
```
bertramakers, line 77:
```suggestion
```json
```

(Also in other codeblocks with JSON)

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+{
+  "audienceType": "education"
```
bertramakers, line 79: This is not 100% correct, the `audienceType` property needs to be nested inside an `audience` property (I'm not sure why it was designed that way in the past, but I think it had something to do with schema.org examples incorrectly being followed to the letter)

Also maybe expand this example to include the other required properties, so it's clear that this needs to be set when creating an event. (You can omit some values by using `"..."`)

For example:
```
{
  "mainLanguage": "nl",
  "name": {
    "nl": "Voorbeeld van een guided tour event"
  },
  "location": {
    "@id": "... (maybe put a URL of a location on test here)"
  },
  "terms": [{"id": "... (just put any applicable eventtype id here)"}],
  "calendarType": "permanent",
  "audience": {
    "audienceType": "education"
  }
}
```

I see that other fields are also required for school events. So maybe put this example at the bottom under an "example(s)" header and include all fields at once. That will probably give the best overview of how the payload should look like.

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+**How to check if your organization already has an organizer page on Cultuurkuur:**
+
+1. Go to <https://www.cultuurkuur.be/organisaties>
+2. Type in the name of your organization in the search box, e.g. `hetpaleis`
+3. If a match is found, open the detail page of the organization
+4. In the url of the browser you'll find the identifier of the organizer in UiTdatabank:
+
+![](../../assets/images/cultuurkuur-search-organizer.png)
+
+Prefix this value with the host url of the according environment and use this as the value for the `organizer.@id` property in the `POST /events` request:
```
bertramakers, line 96:
```suggestion
Prefix this value with the host url of the according environment and use this as the value for the `organizer.@id` property in the `POST /events` request. For example on production:
```

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+}
+```
+
+> Only in the case the organizer of your event does not already have its own page on Cultuurkuur yet, you can [create a new organizer on Cultuurkuur](https://www.cultuurkuur.be/faq/hoe-voeg-ik-mijn-organisatie-toe-op-cultuurkuur).
+
+### labels
+
+For school events, specific Cultuurkuur-related labels are mandatory. There are 3 different types of Cultuurkuur-related labels:
+
+1. **Target group labels** are used to specify if the if event is for students or for teachers.
```
bertramakers, line 112:
```suggestion
1. **Target group labels** are used to specify whether the event is for students or for teachers.
```

(Small typo with 2x "if", but I'd go with "whether" here anyway)

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+2. **Subject labels** are used to define the learning objective of the school event.
+3. **Education level labels**
```
bertramakers, line 114:
```suggestion
3. **Education level labels** indicate to which education levels (grades) the school event is aimed at, for example "Basis onderwijs"
```

(Copied from below and added an example for illustrative purposes)

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+
+#### Target group labels
+
+Target group labels are used to specify if the if event is for students or for teachers.
```
bertramakers, line 118: Again the 2x "if" here, same suggestion as above.

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+
+> A school event must have **exactly 1** target group label.
```
bertramakers, line 120: I think that in English prose, the general rule is to spell out numbers smaller than 10 (see e.g. https://www.scribendi.com/academy/articles/when_to_spell_out_numbers_in_writing.en.html)

```suggestion
> A school event must have **exactly one** target group label.
```

Also applies to other info blocks below (e.g. "A school event must have **at least 1** subject label.")

projects/uitdatabank/docs/entry-api/school-events.md
```diff
@@ -0,0 +1,356 @@
...
+| Education level               | Label (level 4)                              |
+| :---------------------------- | :------------------------------------------- |
+| Eerste leerjaar A             | `cultuurkuur_1ste-leerjaar-A`                |
+| Eerste leerjaar B             | `cultuurkuur_1ste-leerjaar-B`                |
+| Tweede leerjaar               | `cultuurkuur_2de leerjaar`                   |
+| Beroepsvoorbereidend leerjaar | `	cultuurkuur_beroepsvoorbereidend-leerjaar` |
+
+**Tweede graad secundair**
+
+The following level 4 labels must always be combined with
```
bertramakers, line 301: Sidenote and not blocking this guide from being published: I'm seeing this being repeated a lot and it feels like something that we could implement in UDB3 itself? It feels a bit redundant to ask an integrator to add lvl 1, 2 and 3 labels that we can easily derive ourselves from the level 4 label.

projects/uitdatabank/docs/entry-api/school-events.md
corneelwille: Clarified it with quotes!

projects/uitdatabank/docs/entry-api/school-events.md
corneelwille: Good question / remark! From what I remember the main reason we want integrators to contact us is because we want to make sure they only start implementing this if it's the right choice cost-benefit wise. Many cultural organisations already have an integration with our Entry API for their public events. They often only have a few school events each year, which they enter via www.uitdatabank.be. 

For these organisations it's probably more cost effective to keep entering their school events via www.uitdatabank.be instead of developing & maintaining the very custom integration for their school events (with all the custom logic of the location in consultation with the school, complex use of the labels (that change from time to time), etc.)

An integration with the entry api for school events is only cost effective for very big organisations I think, or when you can share the cost of the integration with other customers / partners (like the partners of Parcours vzw).

Long story short: I'll make rephrase the copy / warning and let Eva have a double check on this.

bertramakers: Okay that definitely makes a lot of sense. If you can rephrase this info block to mention it like that I think we will have a lot more success in actually making them contact us to discuss that first before they implement it.

projects/uitdatabank/docs/entry-api/school-events.md
corneelwille: It was a conscious decision not to add the custom logic for Cultuurkuur in UiTdatabank, because there is too little overlap between the goals of Cultuurkuur and the goals of UiTdatabank. 

If you enter a school event on www.uitdatabank.be a user is notified on the preview page that he has to enrich his event on Cultuurkuur (see `https://www.uitdatabank.be/event/49bf2f7f-9ffb-4ef9-b110-6495b85ca06e/preview` for example). The logic of the labels is mostly handled by the user interface on www.cultuurkuur.be: if you add a level 1 label, you'll be able to select the corresponding level 2, level 3 and level 4 labels.

projects/uitdatabank/docs/entry-api/school-events.md
corneelwille: Good remark! But I'm going to omit that nuance to avoid confusion

projects/uitdatabank/docs/entry-api/school-events.md
corneelwille: 👍 Changed it to "location in consultation with the school"

projects/uitdatabank/docs/entry-api/school-events.md
bertramakers: Okay makes sense. Just to clear, I meant on the API though. So if an integrator would send us an event with a level 4 label but not level 3/2/1 labels, we could technically add those automatically. But that's maybe only a solution to a very small group of integrators and not worth it. At least it is clear from the guide I think what is needed!

corneelwille: Yes, indeed it would be to a very small group of integrators. And it would be hard to maintain I think, since the labels for Cultuurkuur change over time (and thus the logic between the levels) - see my comment about "onderwijsvernieuwing".

bertramakers: Alright!

projects/uitdatabank/docs/entry-api/school-events.md
corneelwille: Very useful remark. Added 3 examples on the bottom of the page now

PR #120, opened 2022-11-28
UiTdatabank III-5106 Document workflow status endpoints

bertramakers:
### Added

- Added documentation for the new workflow status endpoints from https://jira.uitdatabank.be/browse/III-4272

---

Ticket: https://jira.uitdatabank.be/browse/III-5106

### DIFF
--- a/projects/uitdatabank/reference/entry.json
+++ b/projects/uitdatabank/reference/entry.json
@@ -4357,6 +4357,79 @@
         ]
       }
     },
+    "/places/{placeId}/workflow-status": {
+      "parameters": [
+        {
+          "$ref": "#/components/parameters/placeId"
+        }
+      ],
+      "put": {
+        "summary": "workflowStatus - update",
+        "tags": [
+          "Places"
+        ],
+        "responses": {
+          "204": {
+            "description": "No Content. The workflowStatus was successfully updated."
+          },
+          "400": {
+            "description": "Bad Request. Possible error types:\n\n* https://api.publiq.be/probs/body/missing\n* https://api.publiq.be/probs/body/invalid-syntax\n* https://api.publiq.be/probs/body/invalid-data",
+            "content": {
+              "application/problem+json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                },
+                "examples": {
+                  "Invalid data": {
+                    "value": {
+                      "type": "https://api.publiq.be/probs/body/invalid-data",
+                      "title": "Invalid body data",
+                      "status": 400,
+                      "schemaErrors": [
+                        {
+                          "error": "Missing required properties (workflowStatus)",
+                          "jsonPointer": "/"
+                        }
+                      ]
+                    }
+                  }
+                }
+              }
+            }
+          },
+          "401": {
+            "$ref": "#/components/responses/Unauthorized"
+          },
+          "403": {
+            "$ref": "#/components/responses/Forbidden"
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound"
+          }
+        },
+        "operationId": "place-workflow-status-put",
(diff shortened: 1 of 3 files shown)


projects/uitdatabank/reference/entry.json
```diff
@@ -4284,6 +4357,79 @@
...
+          },
+          "403": {
+            "$ref": "#/components/responses/Forbidden"
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound"
+          }
+        },
+        "operationId": "place-workflow-status-put",
+        "description": "Updates the workflow status of an place. Possible statuses are:\n\n- `DRAFT`: The default status of new places. Places with this status do not appear in online calendars like <https://www.uitinvlaanderen.be> or the search on <https://www.uitdatabank.be>\n- `READY_FOR_VALIDATION`: This status means the place has been published, but not approved yet. Most online calendars will already show it, and it will appear in the search on <https://www.uitdatabank.be>\n- `APPROVED`: The place has been approved by a moderator. It will appear on all online calendars. You cannot set this status unless you have moderation permissions.\n- `REJECTED`: The place has been rejected by a moderator. It will not appear on any online calendars. You cannot set this status unless you have moderation permissions.\n- `DELETED`: The place has been deleted. It will not appear on any online calendars.",
```
corneelwille, line 4411: This is not completely correct I think. 
* creation through `/places` endpoint -> `DRAFT` as default
* creation through `/imports/places` endpoint -> `READY_FOR_VALIDATION` as default

projects/uitdatabank/reference/entry.json
```diff
@@ -2614,6 +2614,79 @@
...
+          },
+          "403": {
+            "$ref": "#/components/responses/Forbidden"
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound"
+          }
+        },
+        "operationId": "event-workflow-status-put",
+        "description": "Updates the workflow status of an event. Possible statuses are:\n\n- `DRAFT`: The default status of new events. Events with this status do not appear in online calendars like <https://www.uitinvlaanderen.be> or the search on <https://www.uitdatabank.be>\n- `READY_FOR_VALIDATION`: This status means the event has been published, but not approved yet. Most online calendars will already show it, and it will appear in the search on <https://www.uitdatabank.be>\n- `APPROVED`: The event has been approved by a moderator. It will appear on all online calendars. You cannot set this status unless you have moderation permissions.\n- `REJECTED`: The event has been rejected by a moderator. It will not appear on any online calendars. You cannot set this status unless you have moderation permissions.\n- `DELETED`: The event has been deleted. It will not appear on any online calendars.",
```
corneelwille, line 2668: Same remark as for places, the default `DRAFT` is only for the `/events` endpoint and not for the `imports/events` endpoint

projects/uitdatabank/reference/entry.json
```diff
@@ -2614,6 +2614,79 @@
...
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "../models/event-workflowStatus-put.json"
+              },
+              "examples": {}
+            }
+          },
+          "description": "New workflowStatus to set on the event. Depending on the new workflowStatus, other properties may/must be set as well."
```
corneelwille, line 2686: Maybe define which properties must be set as well?

projects/uitdatabank/reference/entry.json
bertramakers: Yes, but that endpoint is not documented anymore. I think it would be confusing to focus here on an undocumented endpoint for existing integrators, since most likely it will be new integrators that will look at the documentation. For one reason because existing integrations don't seem to change every day, and another reason being that existing integrations that use `/imports/places` send all their data in 1 request so they probably wouldn't look at the documentation for this endpoint to update 1 property

projects/uitdatabank/reference/entry.json
bertramakers: I can add it, but it's also mentioned in the descriptions of the possible properties. I've tweaked the descriptions slightly to clarify it. See screenshot below.

Screen Shot 2022-11-28 at 13 53 07

Is this okay or do you think we should also mention it in the description of the complete body?

bertramakers: For reference, you can preview the endpoint here (when logged in): https://docs.publiq.be/docs/uitdatabank/branches/uitdatabank%2FIII-5106-document-workflow-status-endpoints/4c00e3267c774-workflow-status-update

projects/uitdatabank/reference/entry.json
bertramakers: I can add some nuance and make it "The default status of new places (in most cases)." or something like that

corneelwille: As discussed on slack we will provide a guide about how to "publish" your event, since there are many possible ways to do that.

corneelwille (approved):

JonasVHG (approved):

PR #131, opened 2022-12-01
Uitdatabank/search api guides resizing images

corneelwille:
### Added
Documentation about resizing images from UiTdatabank


projects/uitdatabank/docs/search-api/resizing-images.md
```diff
@@ -0,0 +1,203 @@
...
+GET /image/{imageId}?crop=entropy
+```
+
+### format
+
+With the `format` parameter you can convert the image to a different output format.
+
+**Possible values**
+
+See [imgix documentation](https://docs.imgix.com/apis/rendering/format/fm)
```
JonasVHG, line 51: I wouldn't link to an underlying  service. To an integrator it is not imported where and how we host or images.
And maybe even only mention the formats we support as input(jpeg, gif, png)+avif(default on imgx)

JonasVHG (approved):
I'll approve it, cause it is technical OK. 
While anyone can see we use ImgIx via an imagelink, I wouldn't mention it in a guide. 
It is not relevant an end user and might confuse them more, because they might be tempted to use other imgix options.

projects/uitdatabank/docs/search-api/resizing-images.md
bertramakers: @JonasVHG I had the same reflex when reviewing this PR. But I checked with @willaerk and there are very little resizing services left on the market, and we're probably not going to change anytime soon.

There were also already integrators that noticed that we used imgix because they noticed the redirects. And it was also already documented on https://documentatie.uitdatabank.be/content/integreren-met-searchv3/latest/resizing-images.html Those integrators for example asked if they could use imgix directly to skip the redirect, because they assumed it resulted in performance issues (which I doubt).

Given those two factors, we (Kristof/Corneel/me) decided that it would better to just be transparent about it and make it possible to use all imgix parameters on our image URLs. In fact Kristof already had an open ticket for that since 2021: https://jira.uitdatabank.be/browse/OPS-806 

However we will also make it very clear that they should always use our image URLs and not the imgix URLs directly, because the imgix URLs can change (e.g. when we change to a different datastore within imgix).

We're now waiting for Kristof to make it possible to use all imgix parameters, and then Corneel or I will update this guide to:

- Explain that we use imgix but that integrators should use our image URLs to avoid broken links later
- Refer to the imgix documentation for parameters
- Highlight the most useful imgix parameters with some examples (similar to the examples right now)
- Document the "deprecated" parameters and how they map to imgix parameters, for existing integrators. By using the wording "deprecated" we assume that newer ones will avoid these, reducing the chance of confusion because of overlapping parameters (for example if you'd use `width` and `w` at the same time)

corneelwille:
Decided with @willaerk and @JonasVHG to review/merge/publish this docs until the we did the infra work for the imgix parameters

PR #133, opened 2022-12-02
Uitdatabank/document deprecated imports endpoints

bertramakers:
### Added

- Documentation for the deprecated `/imports/...` endpoints, as a reference for existing integrators

### Changed

- Expanded/optimized the descriptions of the endpoints related to creating events/places, as well as the descriptions of their request bodies

---

Best reviewed commit per commit, and/or visually on https://docs.publiq.be/docs/uitdatabank/branches/uitdatabank%252Fdocument-deprecated-imports-endpoints


projects/uitdatabank/models/place-with-write-example.json
```diff
@@ -0,0 +1,143 @@
...
+      "status": {
+        "type": "Unavailable",
+        "reason": {
+          "nl": "Nederlandse reden",
+          "fr": "Raison français",
+          "de": "Deutscher Grund",
+          "en": "English reason"
+        }
+      },
+      "bookingAvailability": {
```
corneelwille, line 48: This is not supported for places, so best to omit this from the example body.

corneelwille (commented):
Very nice and still useful I think. The alerts on top of the page are quite "in your face", but for an endpoint that is deprecated that's okay.
Looks good to me and you're able to merge I think if the remarks I have regarding the `bookingAvailability` are solved.

While reviewing this, I noticed that some things regarding `bookingAvailability` and `status` are not documented quite correctly:

> **bookingAvailability**
> `place.bookingAvailability`
> Indicates whether the place still has tickets or reservations available. Currently only contains a type that can be Available or Unavailable, but can later be expanded with more detailed info.

☝️ This is incorrect, the bookingAvailability of a place can not be set. Currently a place always has the `bookingAvailability` `Available`.
(Also, see my remark on the place-post.json) 

> **status**
> `place.status`
> Indicates if the [place](https://docs.publiq.be/docs/uitdatabank/branches/uitdatabank%2Fdocument-deprecated-imports-endpoints/place.json) is still open for visitations or not.
>
> Examples:
{"type":"Available"}
{"type":"TemporarilyUnavailable","reason":{"nl":"Uitgesteld wegens covid-19","en":"Postponed due to covid-19"}}
{"type":"Unavailable","reason":{"nl":"Permanent gesloten"}}

☝️ Looks like the status reason from events is taken. Best to correct this to something like "Tijdelijk gesloten wegens verbouwingswerken"

For the corrections in the models I'll create a separate ticket

projects/uitdatabank/models/place-with-write-example.json
bertramakers: Thanks @corneelwille , you are correct. But I've taken these examples / descriptions from previously existing endpoints, and I've even indirectly referenced them to avoid duplicates of those models/examples. So I'd suggest that we merge this and then fix those descriptions / examples in another PR, to avoid this PR remaining open due to unrelated issues that are not caused by this PR in itself.

corneelwille (approved):

PR #134, opened 2022-12-05
UiTdatabank: Split reuse guide into two

bertramakers:
### Changed

- The "Reusing existing places & organizers" guide is now split into "Finding and reusing places" and "Finding and reusing organizers". It is no longer visible in the sidebar, but if someone has bookmarked the URL or has it in an email etc they will see a page with an info message to go to one of those two pages.
- The EntryAPI guides are now grouped in "Events", "Places" and "Organizers"

### Fixed

- Minor spelling fixes in the reuse guides


projects/uitdatabank/docs/entry-api/reuse-places-organizers.md
```diff
@@ -1,304 +1,10 @@
...
-      }
-   ]
-}
-```
-
-If the organizer already exists (like in the example above), the URL of that organizer (`https://io.uitdatabank.be/organizers/e83edb7f-6f23-47cd-a730-66fd7687f116` in this case) must be used as `organizer.@id` in the `POST /events` request of the event.
-
-Only when no results are found a new organizer can be created in UiTdatabank. When doing so, the UiTdatabank API will return a response containing the URL of the organizer. This organizer URL must be used then in the `POST /events` request of the event.
+<!-- 
+  This page is not listed in the sidebar, but we should keep it for anyone that has the
```
corneelwille, line 4: I don't think this link is already used by so many external parties? Maybe just link to the new guide?

bertramakers: I've linked to it already a couple of weeks ago in an email to an integrator that had questions about this

bertramakers: And it's easy to have a "forward" like this I think? Also a good example for future guides that get split up. (For simple moves we can do a redirect in Stoplight)

projects/uitdatabank/toc.json
```diff
@@ -25,28 +25,52 @@
...
-          "uri": "/docs/entry-api/reuse-places-organizers.md",
-          "slug": "entry-api/reuse-places-organizers"
+          "type": "group",
+          "title": "Events",
+          "items": [
+            {
+              "type": "item",
+              "title": "School events",
+              "uri": "/docs/entry-api/events/school-events.md",
+              "slug": "entry-api/events/school-events"
```
bertramakers, line 39: I will add a redirect in Stoplight to handle this slug change, so the old slug keeps working

projects/uitdatabank/docs/entry-api/reuse-places-organizers.md
corneelwille: But now we introduce a forward page because 1 email is sent to an integrator?
Also every new integrator that is now on the "Requirements before going live" page, has to click twice to go to the correct reuse guide: one time on the link in the "Requirements before going live" page and one time on the "forward page"

bertramakers: Yes I just noticed the "Requirements before going live" page, I overlooked that. Of course we should link to the separate guides there.

For the link in the email: I just think it's best practice in general, a good example for anyone else who does something like this in the future when the docs are more heavily used, and it's not hard? It's already in place like this and no one will unnecessarily see it when I update the links in our other guides to link to the separate guides

bertramakers: I've fixed the links in other guides to the old page in https://github.com/cultuurnet/apidocs/pull/134/commits/f1d380f2746db1ff293ac3da8e4bbefc38eb77dc

corneelwille: Okay, thanks 👍 
I have nothing against a forward page, but with the link in the "Requirements before going live" page to the forward page it felt like the solution for one integrator (with link) introduced a problem for all other integrators (without link)

corneelwille (approved):

projects/uitdatabank/docs/entry-api/reuse-places-organizers.md
bertramakers: Yep I understand now, that makes sense but an oversight on my end :)

PR #150, opened 2022-12-16
Uitdatabank - Add documentation for GET `/user`

steftrenson:
### Added
- Documentation for GET `/user`

---

Ticket: https://jira.uitdatabank.be/browse/III-5101

### DIFF
--- a/projects/uitdatabank/reference/entry.json
+++ b/projects/uitdatabank/reference/entry.json
@@ -7470,6 +7470,18 @@
                       "type": "string",
                       "example": "username",
                       "description": "The username of the user."
+                    },
+                    "id": {
+                      "type": "string",
+                      "example": "c8e6ff82-8b30-4295-b937-ab2f4f6ab4bf",
+                      "description": "Deprecated, use `uuid` instead.",
+                      "deprecated": true
+                    },
+                    "nick": {
+                      "type": "string",
+                      "example": "username",
+                      "description": "Deprecated, use `username` instead.",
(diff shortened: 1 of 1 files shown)


projects/uitdatabank/reference/entry.json
```diff
@@ -7438,6 +7438,71 @@
...
+                    "email": {
+                      "type": "string",
+                      "example": "info@publiq.be",
+                      "description": "The email address of the user."
+                    },
+                    "username": {
+                      "type": "string",
+                      "example": "username",
+                      "description": "The username of the user."
+                    }
```
bertramakers, line 7473: Someone who calls this endpoint will also get the old `id` and `nick` properties. To avoid confusion, I'd suggest to also document these but with [`"deprecated": true`](https://json-schema.org/draft/2020-12/json-schema-validation.html#section-9.3). That way they will get a deprecated flag in the docs and it will be clear that it's best to use the other ones instead. In the description you can specify which alternatives to use (`uuid` and `username` respectively)

projects/uitdatabank/reference/entry.json
```diff
@@ -7438,6 +7438,71 @@
           "News articles"
         ]
       }
+    },
+    "/user": {
+      "get": {
+        "summary": "user - get",
```
bertramakers, line 7444:
```suggestion
        "summary": "current user - get",
```

bertramakers (approved):

PR #164, opened 2023-01-06
API-110 Fix documentation of the logout flow

bertramakers:
### Fixed

- We should not recommend using `prompt=login` to avoid implementing the logout flow on the authorization server, as it disables SSO. Instead the integrator should always redirect to the `/logout` URL on the Auth0 server when logging out, and only use `prompt=login` if they explicitly want to disable SSO.

---

Ticket: https://jira.uitdatabank.be/browse/API-110

Related: https://jira.uitdatabank.be/browse/PPF-66


projects/authentication/docs/user-access-token.md
```diff
@@ -303,7 +304,7 @@ The following table gives an overview of all the required/optional parameters:
...
-| `prompt`                | `none` (default) or `login`                                                                                                                                                                          | Optional                                                              | If set to `login`, the user will always be asked to login even if they still have an active session on the authorization server. Useful to force the user to login again, which also makes it possible to switch to another account as before.                   |
+| `prompt`                | `login` or omitted (default)                                                                                                                                                                          | Optional                                                              | If set to `login`, the user will always be asked to login even if they still have an active session on the authorization server.                   |
```
JurgenG, line 307: Dit is veel duidelijker, het leek erop dat de waarde 'none' moest worden ingevoerd (terwijl het niet zo was).

projects/authentication/docs/user-access-token.md
```diff
@@ -429,7 +430,23 @@ The best way to check if a refresh token is expired is to exchange it for an acc
...
+
+> If you do not implement this redirect to the `/logout` URL on the authorization server, the next time a user of your application logs in on the same machine they will be logged in as the user that was previously logged in on your application, as the previous user is still logged in on the authorization server.
+
+<!-- theme: warning -->
+
+> You cannot call the `/logout` endpoint as a regular API endpoint, but you must **redirect** your user's web browser to it. Otherwise the authorization server will not be able to clear the user's session.
+
+## Single sign-on
+
+Single sign-on is enabled by default on the UiTiD authorization server. So if a user logs in first on for example <https://www.uitinvlaanderen.be>, and then navigates to your application in the same web browser and clicks the "login" button, they will automatically be logged in as the same user.
```
JurgenG, line 447: Technically speaking, the same browser session (as e.g. in Firefox you can have different containers). But I'm assuming people using containers will be aware of this. So no need to change it for me.

JurgenG (approved):

projects/authentication/docs/user-access-token.md
bertramakers: True, I didn't think of that but it's probably clear what we mean :)

PR #166, opened 2023-01-06
Widgets: Linting™

bertramakers:
### Changed

- Some automatic changes by the linter. The linting check was green already, but these small quirks are still changed by the linter when running the fixer.


bertramakers:
@corneelwille I'm going to merge this without explicit approval since it's just some automatic fixes and no content changes, and it helps to avoid merge conflicts

PR #178, opened 2023-01-24
UiTdatabank: Replace dead link

bertramakers:
### Fixed

- Replaced link to `overheid.vlaanderen.be` about API of the year award (2019) that is suddenly dead


bertramakers:
Going to merge this without explicit approval because this will also cause linting failures in all other branches (even those not related to UiTdatabank)

PR #196, opened 2023-01-31
Fix typo & change structure

corneelwille:
### Changed
- Changed the structure so that it is consistent with the one of UiTdatabank & UiTPAS

### Fixed
- Typo

### DIFF
--- a/projects/widgets/docs/inleiding.md
+++ b/projects/widgets/docs/inleiding.md
@@ -4,13 +4,9 @@
 
 <br>
 
-# UiT-Widgets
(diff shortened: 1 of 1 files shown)


projects/widgets/docs/inleiding.md
```diff
@@ -4,13 +4,9 @@
 
 <br>
```
bertramakers, line 5: Btw this `<br>` should not be necessary (it adds some extra space between the logo and the header underneath but it also looks fine without it)

bertramakers (approved):

PR #200, opened 2023-02-03
III-5108 Add guide about publishing an event

bertramakers:
### Added

- Guide about publishing an event

---

Ticket: https://jira.uitdatabank.be/browse/III-5108

### DIFF
--- a/projects/uitdatabank/docs/entry-api/events/publish.md
+++ b/projects/uitdatabank/docs/entry-api/events/publish.md
@@ -1,12 +1,130 @@
-Coming soon.
+# Publishing an event
 
-<!--
-  @todo
-  Explain:
-  - The high-level moderation workflow (draft > ready_for_validation > accepted/rejected > deleted) => Maybe link to a separate guide for this?
-  - How to publish when creating
-  - How to publish when updating or via separate endpoint
-  - How to publish on a defined datetime in the future (using availableFrom)
-  - Explain that the event will be published until its endDate, or start date in case of "lessenreeks" (availableTo)
-  - Permissions: Who can publish an event? Who can approve/reject an event?
--->
+When [creating a new event](./create.md), it is by default stored in a `DRAFT` state in [UiTdatabank](https://www.uitdatabank.be). 
+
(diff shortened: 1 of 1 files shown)


corneelwille (approved):

projects/uitdatabank/docs/entry-api/events/publish.md
```diff
@@ -1,12 +1,130 @@
...
-  - The high-level moderation workflow (draft > ready_for_validation > accepted/rejected > deleted) => Maybe link to a separate guide for this?
-  - How to publish when creating
-  - How to publish when updating or via separate endpoint
-  - How to publish on a defined datetime in the future (using availableFrom)
-  - Explain that the event will be published until its endDate, or start date in case of "lessenreeks" (availableTo)
-  - Permissions: Who can publish an event? Who can approve/reject an event?
--->
+# Publishing an event
+
+When [creating a new event](./create.md), it is by default stored in a `DRAFT` state in [UiTdatabank](https://www.uitdatabank.be).
```
corneelwille, line 3: There is one exception / method to publish your event that is not explained on this page: events that are created through the (deprecated) endpoint `/imports/events` will also automatically have the workflowStatus `READY_FOR_VALIDATION`. 

Not sure if we need to add documentation about that (because we don't want to promote it), but without documentation it might become confusing over time.

corneelwille: Maybe we can add an `Exceptions` section where we can list all exceptions / specific cases on one page, so that 
- we have documentation about it 
- these exceptions don't distract the attention of the important documentation for the majority of the integrators

corneelwille (approved):

projects/uitdatabank/docs/entry-api/events/publish.md
bertramakers: Right, I forgot about that. I'll try to work it in in a way that is not too distracting from the more common methods

bertramakers: I tried to incorporate it in the current structure of possible scenario's in which you can publish an event

PR #202, opened 2023-02-06
III-5108 Add guide about cancelling/postponing an event

bertramakers:
### Added

- Guide about cancelling or postponing an event

---

Ticket: https://jira.uitdatabank.be/browse/III-5108


projects/uitdatabank/docs/entry-api/events/status.md
```diff
@@ -1,8 +1,141 @@
...
--->
+# Cancelling or postponing an event
+
+When [creating](./create.md) or (usually) when [updating](./update.md) your event, you can mark it as postponed or cancelled to indicate to attendees that it will be moved to another date or not take place at all.
+
+Cancelled events automatically get an *"(geannuleerd)"* label next to their calendar info in online calendars. For example, on UiTinVlaanderen:
+
+<!-- focus: false -->
+
+![Screenshot of a summary of the event "Boshyacintenfestival" on UiTinVlaanderen, as an example of the "(Geannuleerd)" label](../../../assets/images/event-cancelled.png)
```
corneelwille, line 9: I think it is better to use an image / example 
* that has only one status (so not in the title and no duplicate status by UiTinVlaanderen)
* that has a status reason

Example:
Screenshot 2023-02-10 at 13 13 09

(The copy underneath the image is still relevant)

projects/uitdatabank/docs/entry-api/events/status.md
```diff
@@ -1,8 +1,141 @@
...
+```
+
+Note that you may omit the `status` property on the top level, as it will automatically be set based on the `status` of each `subEvent`. If you include it anyway, it will be ignored.
+
+The top-level `status` is determined as follows:
+
+* When at least one `subEvent` has its `status.type` set to `Available`, the top-level `status.type` will automatically be set to `Available` since there is at least one date that the event is still happening as scheduled. 
+* When all subEvents have their `status.type` set to `Unavailable`, the top-level `status.type` will also be set to `Unavailable` since this means that all dates are cancelled.
+* In case of `calendarType` `single`, the top-level `status` will automatically get the same `reason` as the one in the `subEvent` property. In case of `calendarType` `multiple`, the top-level `status` will never contain a `reason` because the reasons can be different from date to date.
+
```
corneelwille, line 114:
```suggestion

Alternatively, you can set the top-level `status` by using the separate [`PUT /events/{eventId}/status`](/reference/entry.json/paths/~1events~1{eventId}~1status/put) endpoint. This endpoint is the equivalent of setting the same `status` on every `subEvent` of the event when updating it in its entirety via [`PUT /events/{eventId}`](/reference/entry.json/paths/~1events~1{eventId}/put).
```

projects/uitdatabank/docs/entry-api/events/status.md
```diff
@@ -1,8 +1,141 @@
...
-  @todo 
-  Explain how to postpone/cancel an event, and optionally a reason.
-  Maybe show screenshot(s) of how postponed/cancelled events look on UiTinVlaanderen (as an example).
-  Also mention that a new event should be created for the new date(s) when postponed.
--->
+# Cancelling or postponing an event
+
+When [creating](./create.md) or (usually) when [updating](./update.md) your event, you can mark it as postponed or cancelled to indicate to attendees that it will be moved to another date or not take place at all.
+
+Cancelled events automatically get an *"(geannuleerd)"* label next to their calendar info in online calendars. For example, on UiTinVlaanderen:
```
corneelwille, line 5: Maybe useful to link to the calendar summary guide (https://docs.publiq.be/docs/uitdatabank/search-api/calendar-summaries) here?

projects/uitdatabank/docs/entry-api/events/status.md
```diff
@@ -1,8 +1,141 @@
...
+```
+
+## Cancelling an event
+
+### With calendarType "single" or "multiple"
+
+To cancel an event with calendarType `single` or `multiple`, update the `status.type` property of the `subEvent` objects that should be cancelled and set it to `Unavailable`.
+
+Optionally, you can include a `reason` property to explain why the event is cancelled on that date. The `reason` property must be an object with language codes like `name`.
+
```
corneelwille, line 79: The `reason` can only be combined with the statuses `Unavailable` and `temporarilyUnavailable`. When setting a (sub)Event back to `Available`, the reason(s) will be automatically removed.

projects/uitdatabank/docs/entry-api/events/status.md
```diff
@@ -1,8 +1,141 @@
...
+        }
+      }
+    }
+  ]
+}
+```
+
+Note that you may omit the `status` property on the top level, as it will automatically be set based on the `status` of each `subEvent`. If you include it anyway, it will be ignored.
+
+The top-level `status` is determined as follows:
```
corneelwille, line 109: The logic for the derived `status` (3 possible statuses) is a bit more complex than the one for `bookingAvailability` (which has 2 possible statuses). See suggestion below.

projects/uitdatabank/docs/entry-api/events/status.md
```diff
@@ -1,8 +1,141 @@
...
+
+* When at least one `subEvent` has its `status.type` set to `Available`, the top-level `status.type` will automatically be set to `Available` since there is at least one date that the event is still happening as scheduled. 
+* When all subEvents have their `status.type` set to `Unavailable`, the top-level `status.type` will also be set to `Unavailable` since this means that all dates are cancelled.
```
corneelwille, line 112:
```suggestion
* When all subEvents have their `status.type` set to `Unavailable`, the top-level `status.type` will also be set to `Unavailable` since this means that all dates are cancelled.
* When all subEvents have their `status.type` set to `temporarilyUnavailable`, the top-level `status.type` will also be set to `temporarilyUnavailable` since this means that all dates are postponed.
* When the subEvents have different statuses, but one or more subEvents have the `status.type` set to `temporarilyUnavailable`, and none to `Available`, the top-level `status.type` will automatically be set to `temporarilyUnavailable` since there is at least one date that the event still takes places, but on a later date.
```

corneelwille: Tbh this becomes a lot of text, maybe a table makes it more clear:


| status.type subEvents | Derived status.type on top event level |
| :-- | :-- |
| At least one `Available` | `Available` |
| All `Unavailable` | `Unavailable` |
| All `temporarilyUnavailable` | `temporarilyUnavailable` |
| Mixed, but one or more `temporarilyUnavailable` and none `Available` | `temporarilyUnavailable` |

projects/uitdatabank/docs/entry-api/events/status.md
corneelwille, line 79:
```suggestion
The `reason` can only be combined with the statuses `Unavailable` and `temporarilyUnavailable`. When setting a (sub)Event back to `Available`, the reason(s) will be automatically removed.
```

projects/uitdatabank/docs/entry-api/events/status.md
bertramakers: Okay, can you send me a link to that event? Because the screenshot seems to be of a low resolution (looking on a Macbook Pro's built-in screen).

projects/uitdatabank/docs/entry-api/events/status.md
bertramakers: Good addition. But I think the casing should be `TemporarilyUnavailable` (capitalized)? If you can correct that in your suggestion I will apply it.

projects/uitdatabank/docs/entry-api/events/status.md
bertramakers: I think this is correct as well and a bit shorter?

```suggestion
* Otherwise, when at least one `subEvent` has its `status.type` set to `TemporarilyAvailable`, the top-level `status.type` will automatically be set to `TemporarilyAvailable` since there is still at least one date that the event still takes place, but on a later date.
* Otherwise, the top-level `status.type` will automatically be set to `Unavailable` since every `subEvent` will have its `status.type` set to `Unavailable` which means that the event is cancelled completely.
```

projects/uitdatabank/docs/entry-api/events/status.md
corneelwille: Here it is: https://www.uitinvlaanderen.be/agenda/e/wereldkeuken/ac844f27-50f1-4867-be9a-c5bad2538e47

projects/uitdatabank/docs/entry-api/events/status.md
corneelwille: Done

projects/uitdatabank/docs/entry-api/events/status.md
corneelwille: I have no preference.

projects/uitdatabank/docs/entry-api/events/status.md
bertramakers: Done!

projects/uitdatabank/docs/entry-api/events/status.md
bertramakers: I used the screenshot from your comment above, because the tooltip is now displaying very small ([UIT-1696](https://jira.uitdatabank.be/browse/UIT-1696))

corneelwille (approved):

PR #204, opened 2023-02-06
III-5214 contributors endpoint

JonasVHG:
### Added

- `contributors` `PUT` endpoint for `events`, `places` & `organizers`

### Todo in following PR
- `GET` request for `contributors`

---

Ticket: https://jira.uitdatabank.be/browse/III-5214


projects/uitdatabank/reference/entry.json
```diff
@@ -1548,6 +1548,80 @@
...
+                "Example": {
+                  "value": [
+                    "info@publiq.be",
+                    "vragen@publiq.be"
+                  ]
+                }
+              }
+            }
+          },
+          "description": "New list of contributors of the event. Previous contributors of the event but not included in this list will be removed from the contributors of the event."
```
bertramakers, line 1621:
```suggestion
          "description": "New list of contributors of the event. Previous contributors of the event that are not included in this list will be removed."
```

Small typo (the "but" is not correct here I think) and suggestion to make it a bit shorter

projects/uitdatabank/reference/entry.json
```diff
@@ -4669,6 +4743,80 @@
...
+                "Example": {
+                  "value": [
+                    "info@publiq.be",
+                    "vragen@publiq.be"
+                  ]
+                }
+              }
+            }
+          },
+          "description": "New list of contributors of the place. Previous contributors of the place but not included in this list will be removed from the contributors of the place."
```
bertramakers, line 4816:
```suggestion
          "description": "New list of contributors of the place. Previous contributors of the place that are not included in this list will be removed."
```

projects/uitdatabank/reference/entry.json
```diff
@@ -6656,6 +6804,80 @@
...
+                "Example": {
+                  "value": [
+                    "info@publiq.be",
+                    "vragen@publiq.be"
+                  ]
+                }
+              }
+            }
+          },
+          "description": "New list of contributors of the organizer. Previous contributors of the organizer but not included in this list will be removed from the contributors of the organizer."
```
bertramakers, line 6877:
```suggestion
          "description": "New list of contributors of the organizer. Previous contributors of the organizer that are not included in this list will be removed."
```

bertramakers (approved):
Very nice, I only noticed a minor typo in a description (repeated 3x because the endpoints are similar). Approving already since the rest looks great. 👌

projects/uitdatabank/models/common-contributors.json
```diff
@@ -0,0 +1,17 @@
+{
+  "title": "common-contributors",
+  "type": "array",
+  "minItems": 1,
+  "uniqueItems": true,
+  "description": "An array of emails of users who need to have edit rights.",
+  "items": {
+    "type": "string",
+    "format": "email"
```
bertramakers, line 9: After [seeing here](https://github.com/cultuurnet/udb3-backend/pull/1253/files#diff-689482cf065520db8fab34b8dc4c31cf5cf0d726e308cadcdb2e365264e31473R18) that the emails can be at most 255 characters long, I think we should add a `"maxLength": "255"` here. (We probably also need to add that in other existing schemas as well, but for new ones I would try to include DB constraints like that already)

bertramakers (changes requested):
(See https://github.com/cultuurnet/apidocs/pull/204/files#r1099964099)

bertramakers (approved):

PR #226, opened 2023-03-29
UPS-4099 added permissions to GET /permissions

erwin1:
### Added

- added permissions array to GET /permissions response

---

Ticket: https://jira.uitdatabank.be/browse/UPS-4099


bertramakers (approved):
I only have a minor suggestion about 1 permission name, aside from that this looks good. 👍 Once we actually return these permission names we cannot change them anymore, so if you agree with my suggestion it's best to edit it now. Otherwise okay to leave it like it is.

PR #228, opened 2023-03-30
UPS-4377 uitpas errors page

erwin1:
### Added

- uitpas errors page

---

Ticket: https://jira.uitdatabank.be/browse/UPS-4377


projects/uitpas/docs/errors.md
```diff
@@ -0,0 +1,341 @@
...
+
+The request you are trying to perform cannot complete because the provided distribution key cannot be found.
+
+## invalid-distributionkey
+
+* **Complete type:** `https://api.publiq.be/probs/uitpas/invalid-distributionkey`
+* **Title**: `Invalid distribution key`
+* **Status**: `500`
+
+The request you are trying to perform cannot complete because the no valid distribution key can be found.
```
bertramakers, line 101:
```suggestion
The request you are trying to perform cannot complete because no valid distribution key can be found.
```

projects/uitpas/docs/errors.md
```diff
@@ -0,0 +1,341 @@
...
+* **Title**: `Distribution key not found`
+* **Status**: `400`
+
+The request you are trying to perform cannot complete because the provided distribution key cannot be found.
+
+## invalid-distributionkey
+
+* **Complete type:** `https://api.publiq.be/probs/uitpas/invalid-distributionkey`
+* **Title**: `Invalid distribution key`
+* **Status**: `500`
```
bertramakers, line 99: Should this not be a `400`?

projects/uitpas/docs/errors.md
```diff
@@ -0,0 +1,341 @@
...
+* **Title**: `INSZ in token required`
+* **Status**: `400`
+
+The request you are trying to perform cannot complete because this request needs a token including an `https://publiq.be/rrn` claim.
+
+## forbidden-underaged
+
+* **Complete type:** `https://api.publiq.be/probs/uitpas/forbidden-underaged`
+* **Title**: `Forbidden Underaged`
+* **Status**: `400`
```
bertramakers, line 219: `403` might be better for Forbidden, or even `451 (Unavailable For Legal Reasons)`?

projects/uitpas/docs/errors.md
```diff
@@ -0,0 +1,341 @@
...
+
+The request you are trying to perform cannot complete because the checkin is currently not allowed.
+
+## rate-limited
+
+* **Complete type:** `https://api.publiq.be/probs/uitpas/rate-limited`
+* **Title**: `Rate limited`
+* **Status**: `429`
+
+The request you are trying to perform cannot complete because it was performed too many times in a given period.
```
bertramakers, line 341: This seems like something that might be better to add a generic error for in the future but okay to leave as-is for now

bertramakers (commented):
Some general suggestions, but take or leave:

- Sorting the errors alphabetically might make it easier to find the one you're looking for (sadly we cannot include the error type as a fragment (`#`) in redirects in Stoplight, so we cannot redirect to the specific error - only the page where it is documented on like this one)
- Where possible, adding some examples _why_ something is not possible can be useful. For example `Event already has ticket sales` -> Why can you not do certain actions when an event already has ticket sales?
- Where possible, adding info how to fix the error could be useful. For example "invalid-uitpas-number" -> What does a valid UiTPAS number look like?

But that can also be optimizations that @brusselsregular and/or @JurgenG can do later so not necessarily required for this PR.

projects/uitpas/docs/errors.md
erwin1: could be an issue for backward compatibility - although clients probably should act on the HTTP code, not the error type in this case

projects/uitpas/docs/errors.md
erwin1: I don't like 403 for this case.
I'm not against 451 but not sure if that is the intended use

projects/uitpas/docs/errors.md
erwin1: changing to 400

erwin1:
> Sorting the errors alphabetically
done

> Event already has ticket sales
this is a special case and I think we can remove this check and error, but I'd like to take it in a separate issue

> What does a valid UiTPAS number look like?
I'll leave this for someone else to do a pass on the descriptions. Although I would be careful documenting 'too much' - we don't want clients to do their own uitpas number validation.

projects/uitpas/docs/errors.md
bertramakers: Also okay to leave it like this for me, it was just a minor suggestion

bertramakers:
> we don't want clients to do their own uitpas number validation.

No true, maybe that example was not very good. But in some cases it may be possible to document how to fix the error. It was more of a general suggestion, not very specific to that error.

projects/uitpas/docs/errors.md
bertramakers: @erwin1 I think this typo is still present after your recent changes. Not sure if you can still automatically apply the suggestion after the new sorting though.

bertramakers (approved):

PR #229, opened 2023-03-30
UPS-4382 new sort proposal

erwin1:
### Changed

- introduced new sort parameter

---

Ticket: https://jira.uitdatabank.be/browse/UPS-4382

### DIFF
--- a/projects/api-design-guidelines/docs/sorting.md
+++ b/projects/api-design-guidelines/docs/sorting.md
@@ -3,11 +3,18 @@
 Endpoints that support sorting based on one or more field names and order(s) must use a `sort` query parameter with the following syntax:
 
 ```
-?sort[field_name_1]=desc&sort[field_name_2]=asc
+?sort=[-]field_name_1[,[-]field_name_2]...
 ```
 
-In the example above, the results would first be sorted by `field_name_1` in a `desc` order, and then by `field_name_2` in an `asc` order.
+Example
+```
+?sort=name,-creationDate
+```
+
+In the example above, the results would first be sorted by `name` in ascending order, and then by `creationDate` in an descending order.
+
+The endpoints using a sort parameter, must clearly document the allowed sort field values.
 
 ## Case sensitivity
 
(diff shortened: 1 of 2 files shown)


projects/uitpas/reference/uitpas.json
```diff
@@ -3651,6 +3651,15 @@
             "style": "form",
             "explode": true
           },
+          {
+            "schema": {
+              "type": "string"
+            },
+            "in": "query",
+            "name": "sort",
+            "description": "Sorts the rewards in a specific order. Possible values are: `name`, `creationdate`, `redeemCount`, `featured`, `points`. Multiple values can be combined in a comma separated string so the rewards are first sorted by the first field, then the second and so on. By default the sort uses ascending order. Descending order can be specified by including a `-` sign before the field that needs to be ordered descending. e.g. use `-redeemCount,points` to sort descending on redeemCount, and next ascending on `points`.",
```
bertramakers, line 3660: Description is pretty clear but I feel like an `example` could also be useful for this parameter. See https://swagger.io/docs/specification/describing-parameters/ > "Parameter Examples"

bertramakers (approved):
I only have a small suggestion documentation-wise but otherwise this looks good 👍

projects/api-design-guidelines/docs/sorting.md
```diff
@@ -3,11 +3,18 @@
 Endpoints that support sorting based on one or more field names and order(s) must use a `sort` query parameter with the following syntax:
 
 ```
-?sort[field_name_1]=desc&sort[field_name_2]=asc
```
bertramakers, line 6: I would keep this under a separate header at the end, for example "Deprecated syntax" and mention that this was the old way but that it was changed for the new method due to issues with the order not always being easy to determine in various programming languages. It can also mention that APIs that implemented the old sort method should keep doing so for BC, but should also implement the new method. (In reality probably Search API, maybe some other small "search" endpoints in Entry API or others)

bertramakers (approved):
Thanks for the extra additions! 🙏

bertramakers:
@erwin1 I will publish the API design guidelines so that this change is also pushed to Stoplight

PR #232, opened 2023-04-03
UPS-4375 document missing errors

erwin1:
### Fixed

- documented missing errors in get redeem status

---

Ticket: https://jira.uitdatabank.be/browse/UPS-4375

### DIFF
--- a/projects/uitpas/reference/uitpas.json
+++ b/projects/uitpas/reference/uitpas.json
@@ -4311,7 +4311,7 @@
             }
           },
           "400": {
-            "description": "Bad Request. Possible error types:\n\n* https://api.publiq.be/probs/uitpas/invalid-uitpas-number\n\n",
+            "description": "Bad Request. Possible error types:\n\n* https://api.publiq.be/probs/uitpas/invalid-uitpas-number\n* https://api.publiq.be/probs/uitpas/passholder-not-found\n* https://api.publiq.be/probs/uitpas/social-tariff-expired\n",
             "content": {
               "application/problem+json": {
                 "schema": {
--- a/projects/uitpas/reference/uitpas.json
+++ b/projects/uitpas/reference/uitpas.json
@@ -4271,7 +4271,7 @@
                     },
                     "reason": {
                       "type": "string",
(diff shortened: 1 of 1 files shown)


projects/uitpas/reference/uitpas.json
```diff
@@ -4311,7 +4311,7 @@
             }
           },
           "400": {
-            "description": "Bad Request. Possible error types:\n\n* https://api.publiq.be/probs/uitpas/invalid-uitpas-number\n\n",
+            "description": "Bad Request. Possible error types:\n\n* https://api.publiq.be/probs/uitpas/invalid-uitpas-number\n* https://api.publiq.be/probs/uitpas/passholder-not-found\n* https://api.publiq.be/probs/uitpas/social-tariff-expired\n",
```
bertramakers, line 4314: This specific endpoint can - according to the documentation - also return `redeemable: false` with `reason: SOCIAL_TARIFF_EXPIRED`. But that reason and this 400 error response with the type `https://api.publiq.be/probs/uitpas/social-tariff-expired` seem to be mutually exclusive.

erwin1: I know, but since we already had this reason, I figured it was better to leave it. But if you think it's better to remove it, I can also do that.

bertramakers: I'm fine with leaving the `reason`, but it can never be returned if you instead return a `400` with type `https://api.publiq.be/probs/uitpas/social-tariff-expired`?

bertramakers: What will the API return exactly in this endpoint if a passholder's social tariff is expired?

erwin1: http 400 with the `uitpas/social-tariff-expired` error type - so the reason `SOCIAL_TARIFF_EXPIRED` will be unused unless we would decide later on that expired social tariffs shouldn't be blocked from (almost) all requests.

bertramakers: Hmm then I would leave the `SOCIAL_TARIFF_EXPIRED` reason out of the documentation after all to avoid confusion. As an integrator will see that response documentation first and it's likely to assume that they will implement a check for that instead of the 400 response with `https://api.publiq.be/probs/uitpas/social-tariff-expired` (or even if they see that, they might just get confused)

bertramakers (approved):

PR #236, opened 2023-04-05
III-5448 guide about creating places

corneelwille:
### Added
- Guide about creating places 

---

Ticket: https://jira.uitdatabank.be/browse/III-5448


projects/uitdatabank/docs/entry-api/places/create.md
```diff
@@ -1,12 +1,316 @@
...
+}
+```
+
+### terms
+
+This property contains a list of [taxonomy terms](../../taxonomy-api/terms.md) that categorize the place in one of various types. Specific partners may also add accessibility facilities as terms.
+
+The possible taxonomy terms can be browsed via the [`GET /terms`](/reference/taxonomy.json/paths/~1terms/get) endpoint on the Taxonomy API.
+
+To create a place, at least one term of the domain `eventtype` and a `scope` that contains `places` is required. For example the following term that categorizes a place as a museum:
```
JonasVHG, line 215: This one is a bit confusing, while the `API` apparantly allows multiple `termIds` only 1 is saved.
Reading this I would conclude it would be possible to create a place that could be both a museum and a library.

JonasVHG (commented):
I am not sure what the ideal solution to post with multiple termIds' would be. 
Blocking it in `udb3-backend` would be a breaking change. But I would at least mention that only 1 `eventType` will be saved, when posting a new `Place`.

projects/uitdatabank/docs/entry-api/places/create.md
corneelwille: I made the copy clearer in commit `c8f73d7`

JonasVHG (approved):

PR #237, opened 2023-04-05
uitdatabank/III-5443 Deleting Organizers

JonasVHG:
### Added

- Guide about deleting `organizers`

---

Ticket: https://jira.uitdatabank.be/browse/III-5443

### DIFF
--- a/projects/uitdatabank/docs/entry-api/organizers/delete.md
+++ b/projects/uitdatabank/docs/entry-api/organizers/delete.md
@@ -12,7 +12,7 @@ Keep in mind that deleting an organizer is an irreversible action. If you delete
 
 ## Required permissions
 
-The same permission checks as [updating an organizer](./update.md) are applicable.
+Only users with the permissions `Organisaties beheren` can delete an organizer.
 
 ## Method
 
--- a/projects/uitdatabank/docs/entry-api/organizers/delete.md
+++ b/projects/uitdatabank/docs/entry-api/organizers/delete.md
@@ -2,9 +2,9 @@
 
 In some exceptional cases, you may want to delete your organizer from UiTdatabank. For example, if it has been entered more than once.
 
-However, you **should not** delete your organizer for any of the following reasons:
+However, you **should not** delete your organizer for any of the following reason:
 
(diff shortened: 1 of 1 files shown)


projects/uitdatabank/docs/entry-api/organizers/delete.md
```diff
@@ -1,7 +1,25 @@
...
+
+* It has no current events, UiTinVlaanderen and other online calendars focus on events and will not show your organizer. 
+
+Note that when you delete an organizer, it is not actually deleted but its `workflowStatus` property will be set to `DELETED`. This will hide it from the autocomplete in the [UiTdatabank web application](https://uitdatabank.be). You can learn more about the `workflowStatus` property in the guide [workflowStatus](../shared/workflow-status.md).
+
+Keep in mind that deleting an organizer is an irreversible action. If you deleted an organizer by accident, the best recourse is to create a new one with the same info.
+
+## Required permissions
+
+Only users with the permissions `Organisaties beheren` can delete an organizer.
```
corneelwille, line 15: This section is missing some information / is not complete. In the following other cases it is possible to delete an organizer:
* you are the creator of the organizer
* you are a contributor of the organizer
* the user _or client that your access token_ belongs to has been given extra permissions to make changes to the organizer

Here you can find a good starting point to add the information in this section: https://docs.publiq.be/docs/uitdatabank/entry-api/events/updating-an-event#permissions

corneelwille (changes requested):

projects/uitdatabank/docs/entry-api/organizers/delete.md
JonasVHG: That's not how it is in the code, an ordinary user cannot delete his/her own organizer. I've checked this with Postman. So either we have to change that in the code or keep the documentation as it is.

corneelwille: My mistake, I only thought it was not possible via the UI but was possible through the API. It feels weird to me that you cannot delete an organizer you created yourself but there's probably a reason why it is implemented the way it is.

In the meantime, the documentation can stay as is

PR #304, opened 2023-09-27
III-5531 Terms of use

corneelwille:
### Added
- terms of use (dutch)

---

Ticket: https://jira.uitdatabank.be/browse/III-5531

### DIFF
--- a/projects/uitdatabank/docs/terms-of-use/terms-of-use.md
+++ b/projects/uitdatabank/docs/terms-of-use/terms-of-use.md
@@ -1,0 +1,107 @@
+# Gebruiksvoorwaarden publicatietools
+
+(dd 20/06/2023)
+
+Onder publicatietools wordt begrepen:
+
+* UiTdatabank API
+* Agendawidgets: agendamodule gehost door publiq
+* Culturefeed Module Suite: een set van agenda-modules gebaseerd op Drupal
+
+De agendawidgets en Culturefeed Modules zijn gebouwd bovenop de UiTdatabank API.
+
+## I. Onvoorwaardelijke aanvaarding van de gebruiksvoorwaarden
+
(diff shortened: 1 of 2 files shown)


projects/uitdatabank/docs/terms-of-use/terms-of-use.md
```diff
@@ -0,0 +1,107 @@
...
+
+(dd 20/06/2023)
+
+Onder publicatietools wordt begrepen:
+
+* UiTdatabank API
+* Agendawidgets: agendamodule gehost door publiq
+* Culturefeed Module Suite: een set van agenda-modules gebaseerd op Drupal
+
+De agendawidgets en Culturefeed Modules zijn gebouwd bovenop de UiTdatabank API.
```
JonasVHG, line 11: (Same comment for line 58): While the Culturefeed Modules are being upgraded by a third party, instead of deprecated as initially planned. There is no mention of Culturefeed in the rest of the docs, so I wonder if it belongs here?(Open question, legally it might be wise to put it here).

JonasVHG (approved):
Just an open question about mentioning Culturefeed. I do not have the legal knowledge to know if it should be here.
Rest looks good, so I'll approve.

projects/uitdatabank/docs/terms-of-use/terms-of-use.md
corneelwille: Yes, the terms of use are definitely outdated. I already wrote a new version (without mentioning culturefeed modules, with information about auth0 being a processor of user data, etc.) but it needs a review of a jurist. So for now I just copied the terms of use available on [documentatie.uitdatabank.be](https://documentatie.uitdatabank.be/content/integratie-op-je-eigen-website/latest/gebruiksvoorwaarden.html). To be continued!

PR #320, opened 2023-10-10
Projects: Fix Linting errors for redirects

JonasVHG:
### Changed

- Replaced all the redirects with their new links, because the linting started to fail on it.

---


corneelwille (approved):
Looks good 👍

PR #415, opened 2024-12-02
Added documentation for GET /ownerships

Anahkiasen:
### Added

- Added documentation for GET /ownerships

---

Ticket: https://jira.publiq.be/browse/III-6351

### DIFF
--- a/projects/uitdatabank/reference/entry.json
+++ b/projects/uitdatabank/reference/entry.json
@@ -8301,7 +8301,93 @@
         ],
         "x-internal": true
       },
-      "parameters": []
+      "parameters": [],
+      "get": {
+        "summary": "ownership - get",
+        "operationId": "get-ownerships",
+        "responses": {
+          "200": {
+            "description": "",
+            "content": {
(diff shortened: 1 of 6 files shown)


projects/uitdatabank/reference/entry.json
```diff
@@ -8301,7 +8301,86 @@
...
+            "$ref": "#/components/responses/Forbidden"
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound"
+          }
+        },
+        "tags": [
+          "Ownerships"
+        ],
+        "description": "Get all or one specific ownership requests of the current user",
```
LucWollants, line 8356: I don't think the current user is important. Also it doesn't return "all" ownerships, it are the ownerships matching the search query.

This makes me wonder if we need a dedicated endpoint `GET /ownerships/search` like `GET /ownerships/suggestions` to make the difference more clear with `GET /ownerships/{ownershipId}

LucWollants (approved):

PR #439, opened 2025-04-17
III-6579 Endpoint municipalities and education levels

grubolsch:
### Added

- Add 2 new internal endpoints, /regions and /educational-levels, meant for the new Cultuurkuur flow

### Related pr
- https://github.com/cultuurnet/udb3-backend/pull/2047

---

Ticket: https://jira.uitdatabank.be/browse/III-6579

### DIFF
--- a/projects/uitdatabank/reference/entry.json
+++ b/projects/uitdatabank/reference/entry.json
@@ -9507,7 +9507,8 @@
             "headers": {
               "content-type": {
                 "schema": {
-                  "type": "string"
+                  "type": "string",
+                  "example": "application/json"
                 },
                 "description": "application/json"
               }
@@ -9529,6 +9530,7 @@
                       },
                       "children": {
                         "type": "array",
+                        "description": "Divisions of this region",
                         "items": {
                           "type": "object",
                           "properties": {
@@ -9542,6 +9544,7 @@
                             },
                             "children": {
                               "type": "array",
+                              "description": "Divisions of this region",
                               "items": {
                                 "type": "object",
                                 "properties": {
@@ -9670,9 +9673,6 @@
         },
         "operationId": "get-regions",
         "x-internal": true,
-        "requestBody": {
-          "content": {}
-        },
         "parameters": [],
         "description": "Returns a hierarchical list of the regions in Flanders and Brussel. Each level includes a label, value with the NIS-code, and optional nested children, allowing integrators to display lists of all regions and municipalities."
       }
@@ -9687,7 +9687,8 @@
             "headers": {
               "content-type": {
                 "schema": {
-                  "type": "string"
+                  "type": "string",
+                  "example": "application/json"
                 },
                 "description": "application/json"
               }
@@ -9709,6 +9710,7 @@
                       },
                       "children": {
                         "type": "array",
+                        "description": "Divisions of this education level",
                         "items": {
                           "type": "object",
                           "properties": {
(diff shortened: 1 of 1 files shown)


LucWollants:
Should we create a `cultuurkuur` group and provide the endpoints `/cultuurkuur/regions` and `cultuurkuur/education-levels`?

grubolsch:
Changed Taxonomy tag to Cultuurkuur tag, and added /cultuurkuur for each new endpoint.

projects/uitdatabank/reference/entry.json
```diff
@@ -9496,6 +9496,320 @@
           }
         ]
       }
+    },
+    "/cultuurkuur/regions": {
+      "get": {
+        "summary": "/cultuurkuur/regions",
```
LucWollants, line 9502: This should be something like `regions - GET`, you can compare this with existing endpoints.

projects/uitdatabank/reference/entry.json
```diff
@@ -9496,6 +9496,320 @@
...
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "array",
+                  "items": {
+                    "type": "object",
+                    "properties": {
+                      "label": {
+                        "type": "string",
+                        "description": "Name region"
```
LucWollants, line 9525: During the meeting with Eva, we discussed that these labels should be translatable

> We will not translate regions and educational levels, but we will make sure the nl field is used

```
label: {
 nl: 'Gewoon basisonderwijs'
},
```

projects/uitdatabank/reference/entry.json
```diff
@@ -9496,6 +9496,320 @@
...
+                      },
+                      "value": {
+                        "type": "string",
+                        "description": "NIS code"
+                      },
+                      "children": {
```
LucWollants, line 9531: We should think about more expressive names, for example `regions` and `subregions`

projects/uitdatabank/reference/entry.json
```diff
@@ -9496,6 +9496,320 @@
...
+        },
+        "operationId": "get-regions",
+        "x-internal": true,
+        "parameters": [],
+        "description": "Returns a hierarchical list of the regions in Flanders and Brussel. Each level includes a label, value with the NIS-code, and optional nested children, allowing integrators to display lists of all regions and municipalities."
+      }
+    },
+    "/cultuurkuur/education-levels": {
+      "get": {
+        "summary": "/cultuurkuur/education-levels",
```
LucWollants, line 9682: Same remark here, something like `educationLevels - GET`, but the URL path `education-levels`

projects/uitdatabank/reference/entry.json
```diff
@@ -9496,6 +9496,320 @@
...
+                              "type": "string",
+                              "description": "NIS code"
+                            },
+                            "children": {
+                              "type": "array",
+                              "description": "Divisions of this region",
+                              "items": {
+                                "type": "object",
+                                "properties": {
+                                  "label": {
```
LucWollants, line 9551: I think this should be `name`

projects/uitdatabank/reference/entry.json
grubolsch: I am open to that, but maybe we should have that discussion with Emma and then change it in a later PR. I think the current way might it easier to write some recursive code in JS, but that is just speculation.

JonasVHG (approved):

PR #453, opened 2025-06-05
III-6656 update school event page

JonasVHG:
### Changed

- `projects/uitdatabank/docs/entry-api/events/create-school.md`: Updated/added info about workingregions, educational levels, priceInfo, eventtypes

---

Ticket: https://jira.publiq.be/browse/III-6656


LucWollants (commented):
Don't we need to mention the education levels like in the previous documentation? Where will an integrator find the education levels he can use?

LucWollants:
> > Don't we need to mention the education levels like in the previous documentation? Where will an integrator find the education levels he can use?
> 
> There is a link to the endpoint(https://io.uitdatabank.be/cultuurkuur/education-levels) in the new docs. Personally I am not in favour of copying that list into the docs, because if some changes will be made in the future, it would require us to also keep the docs in sync. Past experiences, when we had both a list of terms in an endpoint & the docs, have shown that this is easily forgotten. Off course if Eva/CultuurKuur explicitly requests it, but I would argue against duplicated lists that could easily go out-of-sync in the future.

My mistake! The link to the education levels endpoint is perfect!

PR #454, opened 2025-06-10
Update client-access-token.md

JurgenG:
Replacing not-working CURL request with a better alternative

### Changed

- curl instruction in Client Access documentation

---

Ticket: https://jira.uitdatabank.be/browse/III-6603


JonasVHG (approved):

bertramakers:
@JurgenG this call should still work with JSON, even after the move to Keycloak, otherwise all existing integrations would have broken at the time. If the curl example does not work, I'd suggest to fix the specific issue but keep using the JSON content-type. While it's maybe a bit more verbose in curl, it's easier for integrators to implement programmetically. And if we use form data in the curl example, developers might think we only support form data.

PR #461, opened 2025-07-10
III-6716 Clarify Docs about searching DRAFTS

JonasVHG:
### Changed

- `default-filter.md`: Clarify documentation about searching for `DRAFT`.

---

Ticket: https://jira.publiq.be/browse/III-6716

### DIFF
--- a/projects/uitdatabank/docs/search-api/filters/default-filters.md
+++ b/projects/uitdatabank/docs/search-api/filters/default-filters.md
@@ -19,9 +19,9 @@ You can either disable these filters individually, or disable them all at once.
 
 > If you search for 1 specific filtered group, the desired results might still be blocked because they are still blocked by another filter.
 > e.g., Offers in `DRAFT` are always unavailable. So if wish to search for `DRAFT`-offers you must also disable `availableFrom` & `availableTo`
-```
+> ```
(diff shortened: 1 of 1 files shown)


grubolsch (approved):

projects/uitdatabank/docs/search-api/filters/default-filters.md
```diff
@@ -14,6 +14,16 @@ You can either disable these filters individually, or disable them all at once.
 
 > If you disable (one of) these filters, you may get results that should not be published in event calendars or are no longer relevant to them.
 
+
+<!-- theme: info -->
+
+> If you search for 1 specific filtered group, the desired results might still be blocked because they are still blocked by another filter.
```
LucWollants, line 20: Is "blocked" the correct term in this context? Should it not be "hidden"?

The AI-generated explanation with the given context

> When performing a search using multiple filters, keep in mind that results for a specific group (such as offers with a particular workflow status) may still be excluded if other filters are active. For example, even if you set the filter to find offers with workflowStatus=DRAFT, those offers will not appear unless you also disable the availableFrom and availableTo filters. This is because offers in DRAFT status are always hidden by these availability filters. To search for DRAFT offers, make sure to remove or bypass conflicting filters that might block your desired results.

PR #508, opened 2026-04-20
III-7136 Add openingHoursClosedDays and openingHoursAdjustedDays to places

grubolsch:
## Summary

- Renames `event-openingHoursClosedDays.json` → `common-openingHoursClosedDays.json` since the schema is identical for events and places (no childcare involved)
- Adds `place-openingHoursAdjustedDays.json` that mirrors the event version but uses `place-openingHours.json` (no childcare support)
- Adds `openingHoursClosedDays` and `openingHoursAdjustedDays` to `place.json` and `place-calendar-put.json`
- Updates `place-with-read-example.json` with example values for both new properties
- Updates `calendar-info.md` to remove "events only" restrictions; clarifies childcare within adjusted opening hours remains events-only

## Test plan

- [ ] Verify `place.json` and `place-calendar-put.json` expose both new properties
- [ ] Verify `event.json` and `event-calendar-put.json` still resolve correctly (now pointing to `common-openingHoursClosedDays.json`)
- [ ] Verify `place-openingHoursAdjustedDays.json` does not allow `childcare` (uses `place-openingHours.json`)
- [ ] Verify `event-openingHoursAdjustedDays.json` still allows `childcare` (uses `event-openingHours.json`)
- [ ] Check docs render correctly in `calendar-info.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
Ticket https://jira.publiq.be/browse/III-7136

### DIFF
--- a/projects/uitdatabank/models/common-openingHoursAdjustedDays.json
+++ b/projects/uitdatabank/models/common-openingHoursAdjustedDays.json
@@ -1,0 +1,78 @@
+{
+  "description": "List of specific date ranges during which custom opening hours override the default schedule. Only applicable for calendarType `periodic` and `permanent`. An empty array clears all previously set adjusted opening hours.",
+  "type": "array",
+  "items": {
+    "type": "object",
+    "properties": {
+      "startDate": {
+        "type": "string",
+        "format": "date",
+        "description": "Start of the adjusted opening hours date range (inclusive), in ISO 8601 date format (YYYY-MM-DD). For periodic events and places, must fall within the main startDate/endDate.",
+        "examples": ["2026-12-21"]
+      },
+      "endDate": {
+        "type": "string",
+        "format": "date",
+        "description": "End of the adjusted opening hours date range (inclusive), in ISO 8601 date format (YYYY-MM-DD). For periodic events and places, this must be on or after startDate, and within the main startDate/endDate. When openingHoursClosedDays overlap with this range, the closed days take precedence.",
+        "examples": ["2027-01-03"]
+      },
+      "description": {
+        "description": "Optional human-readable description of the adjusted opening hours period, as localized text. Maximum 1000 characters per language.",
+        "type": "object",
+        "minProperties": 1,
+        "additionalProperties": false,
+        "properties": {
+          "nl": { "type": "string", "maxLength": 1000, "minLength": 1, "description": "Dutch description." },
+          "fr": { "type": "string", "maxLength": 1000, "minLength": 1, "description": "French description." },
+          "de": { "type": "string", "maxLength": 1000, "minLength": 1, "description": "German description." },
+          "en": { "type": "string", "maxLength": 1000, "minLength": 1, "description": "English description." }
+        }
+      }
+    },
+    "required": ["startDate", "endDate", "openingHours"],
+    "examples": [
+      {
+        "startDate": "2026-12-21",
+        "endDate": "2026-12-30",
+        "description": {
+          "nl": "Kerstvakantie",
+          "fr": "fêtes de Noël"
+        },
+        "openingHours": [
+          {
+            "opens": "13:00",
+            "closes": "15:00",
+            "dayOfWeek": [
+              "friday",
+              "saturday",
+              "sunday"
+            ]
+          }
+        ]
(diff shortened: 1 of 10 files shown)


grubolsch:
@claude

grubolsch:
@claude

projects/uitdatabank/models/place-openingHoursAdjustedDays.json
lucwollants, line 1: I wonder if it is possible to combine the event adjusted days and place adjusted days if only the opening hours are different? Not a blocker, just a thought.

grubolsch: You mean a cross check to check if the opening hours of an event are inside the opening hours of the place?

lucwollants: No, using one `common-openingHoursAdjustedDays.json` that gets embedded in `place-openingHoursAdjustedDays.json` and `event-openingHoursAdjustedDays.json` with the only difference of using `place-openingHours.json` instead of `event-openingHours.json`

This was also remarked by Claude Code, see:
> 2. place-openingHoursAdjustedDays.json code duplication
The file is structurally identical to event-openingHoursAdjustedDays.json except for the $ref in openingHours and the title. As noted, this is intentional. Just flagging it so it's a conscious trade-off rather than an oversight. If the description/item shape ever diverges further, having them separate will pay off.

But not a blocker. This is maybe an optimisation that can be done in a follow-up pull request?

grubolsch:
@claude

projects/uitdatabank/models/place-openingHoursAdjustedDays.json
grubolsch: It is a good idea, so i already did the change

grubolsch:
Ready for review

projects/uitdatabank/models/common-openingHoursAdjustedDays.json
```diff
@@ -0,0 +1,78 @@
...
+        "additionalProperties": false,
+        "properties": {
+          "nl": { "type": "string", "maxLength": 1000, "minLength": 1, "description": "Dutch description." },
+          "fr": { "type": "string", "maxLength": 1000, "minLength": 1, "description": "French description." },
+          "de": { "type": "string", "maxLength": 1000, "minLength": 1, "description": "German description." },
+          "en": { "type": "string", "maxLength": 1000, "minLength": 1, "description": "English description." }
+        }
+      }
+    },
+    "required": ["startDate", "endDate", "openingHours"],
```
lucwollants, line 39: I'm confused here, because `openingHours` is marked as required but seems to be missing from the `properties`? Should this not be added and pointing to `common-openingHours.json`?

projects/uitdatabank/models/common-openingHoursAdjustedDays.json
lucwollants, line 1: Is this file missing a title?

projects/uitdatabank/models/common-openingHoursAdjustedDays.json
```diff
@@ -0,0 +1,78 @@
...
+              "friday",
+              "saturday",
+              "sunday"
+            ]
+          }
+        ]
+      }
+    ]
+  },
+  "examples": [
```
lucwollants, line 62: This seems to be an exact copy of line 33?

projects/uitdatabank/models/common-openingHoursAdjustedDays.json
lucwollants: An existing issue, but it seems that event opening hours examples are missing childcare examples.

projects/uitdatabank/models/common-openingHoursAdjustedDays.json
grubolsch: They’re not duplicates because they show different levels:

items.examples: shows what one item looks like ({ startDate, endDate, ... })
top-level examples: shows the full array ([{ startDate, endDate, ... }]) ( extra [] )

I copied it from common-openingHoursClosedDays.json.

lucwollants (approved):

PR #509, opened 2026-04-20
III-7134 Add capacity to place bookingAvailability

grubolsch:
## Summary

- Adds optional `capacity` property to `place-bookingAvailability.json`, reusing the existing `common-bookingAvailability-capacity.json` component (integer ≥ 0)
- Updates the place read and write examples to include `capacity` in `bookingAvailability`
- Rewrites the stub `places/booking-availability.md` into a proper guide documenting the property, an example, and how to update it

## Ticket
https://jira.publiq.be/browse/III-7134


grubolsch:
@claude

grubolsch:
@claude

lucwollants (approved):
Looks good.

I only wonder if we now also need a top-level booking-availability endpoint for places because we can now set the capacity. I raised this question in Slack and it is outside the scope of this pul request.

PR #519, opened 2026-05-13
HOTFIX: from to are not references in `birthdateRangePut`

JonasVHG:
### Fixed

- `projects/uitdatabank/models/event-birthdateRange-put.json`: `from` & `to` are not references.
- `projects/uitdatabank/docs/entry-api/events/create-boa.md`: Updated slugs.

---


bertramakers (commented):
Can't we just reference the complete `event-birthdateRange.json` schema in this operation? Like for example in:

- https://github.com/cultuurnet/apidocs/blob/main/projects/uitdatabank/models/event-contactPoint-put.json 
- https://github.com/cultuurnet/apidocs/blob/main/projects/uitdatabank/models/event-contributors-put.json
- etc ...

lucwollants (approved):

PR #525, opened 2026-05-28
III-7155 Improve image upload documentation

steftrenson:
Added

  - New ### URL import (application/json) section in the Images guide (shared/images) documenting the contentUrl-based upload method, including required fields and an HTTP example.
  - Example for the application/json request body on POST /images in the OpenAPI spec, so it renders correctly in Stoplight's "Try it" panel.

  Changed

  - Restructured the "Uploading an image" section in the Images guide to clearly present both upload methods (multipart/form-data and application/json) side by side, instead of only describing the binary file upload.
  - Updated the POST /images endpoint description in the OpenAPI spec to mention both upload methods upfront.

  Fixed

  - contentUrl in image-post.json was referencing common-string-uri-readOnly.json, marking it as a read-only property in a request body schema. Switched to common-string-uri.json so the field is correctly treated as writable.

  ---
  Ticket: https://jira.uitdatabank.be/browse/III-7155

---
Reworked images guide can be found [here](https://publiq.stoplight.io/docs/uitdatabank/branches/III-7155-image-upload-documentation/entry-api/shared/images)
Reworked endpoint documentation can be found [here](https://publiq.stoplight.io/docs/uitdatabank/branches/III-7155-image-upload-documentation/entry-api/reference/operations/create-a-image)

### DIFF
--- a/projects/uitdatabank/docs/entry-api/shared/images.md
+++ b/projects/uitdatabank/docs/entry-api/shared/images.md
@@ -26,7 +26,12 @@ When updating an existing [event](../events/update.md), [place](../places/update
 
 Before you can link an image to an event, place or organizer you must first upload it via the [`POST /images`](/reference/entry.json/paths/~1images/post) endpoint.
 
-This endpoint expects data in the `multipart/form-data` content-type, to enable the file being uploaded as binary data.
+Two upload methods are supported:
+
+* **Binary file upload** (`multipart/form-data`): upload an image file directly as binary data.
+* **URL import** (`application/json`): provide a URL pointing to an existing online image that UiTdatabank will import.
+
+### Binary file upload (multipart/form-data)
 
 The following fields are required:
 
@@ -35,7 +40,7 @@ The following fields are required:
 * `copyrightHolder`: The name of the person or entity that holds the copyright on the image. Will be used for attribution.
 * `language`: The language of the image and its description. For example, an image can have some text on it in a specific language.
 
-### HTTP example
+#### HTTP example
 
 The snippet below is an example of how a `multipart/form-data` request with the binary file data and other required fields should look like when sent to Entry API:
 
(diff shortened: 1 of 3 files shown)


projects/uitdatabank/docs/entry-api/shared/images.md
```diff
@@ -226,6 +231,31 @@ var_dump($response);
...
+### URL import (application/json)
+
+When the image is already available online, you can import it by providing its URL instead of uploading binary data. Send the request with `Content-Type: application/json` and the following required fields:
+
+* `contentUrl`: The URL of the existing online image to import.
+* `description`: A description of the image.
+* `copyrightHolder`: The name of the person or entity that holds the copyright on the image. Will be used for attribution.
+* `inLanguage`: The language of the image and its description.
+
+#### HTTP example
```
lucwollants, line 245: Should we include more examples as we did for the binary upload? This approach is much easier, so less convoluted documentation is needed?

steftrenson: Personally, I think the url upload method is much easier than the binary upload and doesn't warrant extra documentation. But happy to add it if you think it'll add value.

steftrenson:
@claude please review.

steftrenson:
@bertramakers Do you think this documentation "change" should be mentioned in the release notes?

bertramakers:
For sure!
I noticed @grubolsch already added it to the release notes he created retro-actively for 2026: https://github.com/cultuurnet/apidocs/pull/528/changes#diff-6e27bbf9553dfcbcc90334b04f4f91627d9a13f115ad8236c37e2b5752f2effeR76
So I don't think you need to do anything extra on your end for the release notes unless you think that bullet point can be improved or it should be added in a more recent release note

lucwollants (approved):

PR #530, opened 2026-06-09
Remove references to internal changes in 2025 release notes

bertramakers:
### Removed

- Release notes about internal ownerships endpoints
- Release notes about internal saved-searches endpoints
- Release notes about internal refactorings (e.g. Address valueobject)
- Release notes about internal optimizations (e.g. caching, crashing projectors, de-duplicated config, ...)
- Release notes about internal labels endpoint
- Release notes about internal user endpoint
- Release notes about internal address endpoint (Bpost proxy)
- Release notes about internal RDF projection
- Release notes about internal Cultuurkuur changes
- Release notes about CLI commands and cronjobs

---

Not 100% sure about some but I double checked per release note if an endpoint was documented or not, and if yes if it was marked as internal or not. If it was not documented I also consider it internal since it's not easily discoverable by integrators (like endpoints marked as internal)

### DIFF
--- a/projects/release-notes/docs/uitdatabank/2025.md
+++ b/projects/release-notes/docs/uitdatabank/2025.md
@@ -2,7 +2,7 @@
 
 ## December
 
-No public-facing changes.
+No public-facing API changes.
 
 ## November
 
@@ -15,9 +15,7 @@ No public-facing changes.
 
 ## October
 
-🛠 **Improvements**
-
-* The "beheeraanvraag" confirmation mail now links to the manage-ownership page instead of the previous (outdated) link.
+No public-facing API changes.
 
 ## September
 
--- a/projects/release-notes/docs/uitdatabank/2025.md
+++ b/projects/release-notes/docs/uitdatabank/2025.md
@@ -63,10 +63,6 @@ No public-facing API changes.
 
 ## February
 
-✨ **New features**
-
-* Description and images of organizers are now included in their RDF projection.
-
 🛠 **Improvements**
 
 * The `/labels` endpoint now enforces a maximum limit of 30 results and uses sensible defaults when called without `start` / `limit`, to mitigate denial-of-service requests.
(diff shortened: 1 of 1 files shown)


projects/release-notes/docs/uitdatabank/2025.md
```diff
@@ -79,73 +51,32 @@
...
-
-* Ownership requests now validate that `ownerId` exists in the system, with a clean error for unknown users.
-* The existing CLI command for updating locality now also supports organizers (in addition to places), to support municipal mergers.
-* New admin CLI command `ownership:add` to add ownerships in bulk without sending notification mails.
-
 ## April
 
 ✨ **New features**
 
-* New backend endpoints serving a static list of municipalities (including their educational levels) and a list of educational levels, to be consumed by Cultuurkuur.
```
bertramakers, line 92: @grubolsch I couldn't find this endpoint in the docs so I considered it internal, but if I overlooked it we can revert this of course. In any case if it's documented it might be best to clarify which endpoint this is about

willaerk (approved):

bertramakers:
FYI for anyone reviewing, feedback/approval is always welcome but I'm also going to wait on feedback from @grubolsch since he initially made the effort to document all of these release notes retro-actively

projects/release-notes/docs/uitdatabank/2025.md
grubolsch: It is internal, and used for the CK frontend changes made last year.

grubolsch:
First of all, thank you for the significant effort that went into this. :-)

I am happy to merge as is, but I do want to flag one point of concern.

When I made the orginal PR, I intentionally left a few entries in each month. My concern is that, without them, the product timeline may give the impression that the product is only updated occasionally, rather than being actively maintained and supported by a (cool) team.

This may ultimately not be an issue, and maybe not the place of release documents to be part of "marketing", but that was the reasoning behind my decision.

bertramakers:
Makes total sense @grubolsch . I also considered that, which is why I added a note saying "No public-facing API changes" to indicate that while there might not be release notes for a given month, it's because we only mention public-facing API changes in these notes. However we can always improve that if you feel like that does not cover it sufficiently. Let's merge this for now and maybe discuss it with Sarah sometime?

PR #534, opened 2026-06-11
III-7233 refactor childrenOnly to a separate boolean

grubolsch:
## Summary

Refactors `childrenOnly` away from the `audienceType` enum into its own dedicated boolean property on the event.

- Removes `childrenOnly` from the `audienceType` enum
- Adds a new optional `childrenOnly` boolean on event (default `false`, not required)
- Adds a new dedicated endpoint `PUT /events/{eventId}/children-only` to update the flag
- Updates `departurePlaces` validation so it now requires `childrenOnly: true` instead of `audienceType: childrenOnly`
- Updates the create-boa guide and event read/write examples accordingly

Omitting `childrenOnly` from the PUT endpoint leaves the existing value unchanged.

## Preview
https://publiq.stoplight.io/docs/uitdatabank/branches/III-7233%2Frefactor-childrenonly-to-a-seperate-boolean/entry-api/reference/operations/update-a-event-children-only


lucwollants (commented):
As discussed through Slack, it makes sense to explain that when `childrenOnly` is `false`, it will be omitted from the projection.

lucwollants (approved):

PR #535, opened 2026-06-11
Update URI regex

grubolsch:
Update the readonly version of common uri to extend common uri so they share the same regex (which was made a little bit stricter).
The old version did have "(readonly)" in the title, I dropped this because this seemed such a low value addition, but I can add it back when requested


lucwollants (commented):
We also have the pattern on https://github.com/cultuurnet/apidocs/blob/main/projects/uitdatabank/reference/entry.json#L9554

Is this a breaking change?

lucwollants (approved):

PR #536, opened 2026-06-12
Restrict bookingAvailability `Unavailable` for permanent/periodic calendars

grubolsch:
## changes

- Based on the feedback of Luc on PR https://github.com/cultuurnet/udb3-backend/pull/2265 I decided to enforce on the json schema the rule that when when `calendarType` is `permanent` or `periodic`, `bookingAvailability.type` cannot be `Unavailable`.
- `Unavailable` remains valid for `single` and `multiple` calendar types, where individual subEvents exist.

## related
https://github.com/cultuurnet/udb3-backend/pull/226

### DIFF
--- a/projects/uitdatabank/models/event-calendar-put.json
+++ b/projects/uitdatabank/models/event-calendar-put.json
@@ -4,6 +4,27 @@
   "description": "Request body to update the calendar information of an event.",
   "allOf": [
     {
+      "if": {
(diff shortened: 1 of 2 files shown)


grubolsch:
One remark here: this is in some way backwords compatible breaking. It would not have worked before, but you did not get an error message.

PR #551, opened 2026-07-08
UPS-6152 added registrationPicture in passholder creation

erwin1:
### Added

- Added registrationPicture in passholder creation

---

Ticket: https://jira.uitdatabank.be/browse/UPS-6152

### DIFF
--- a/projects/uitpas/reference/uitpas.json
+++ b/projects/uitpas/reference/uitpas.json
@@ -11632,6 +11632,15 @@
             "description": "Only used in passholder creation using a `registrationUitpasNumber` with social tariff. Or when updating a passholder with social tariff. During registration, if no `registrationSocialTariffEndDate` is specified and `registrationUitpasNumber` has social tariff, a default date of April 30 of next year is used.",
             "format": "date",
             "writeOnly": true
+          },
+          "registrationPicture": {
+            "type": "string",
+            "description": "Base64 encoded string of the picture of this passholder. Only used in passholder creation.",
+            "format": "byte",
+            "x-stoplight": {
+              "id": "ikgxjklji2guj"
+            },
+            "writeOnly": true
           }
         },
         "required": [


brusselsregular (approved):

projects/uitpas/reference/uitpas.json
```diff
@@ -11632,6 +11632,15 @@
             "description": "Only used in passholder creation using a `registrationUitpasNumber` with social tariff. Or when updating a passholder with social tariff. During registration, if no `registrationSocialTariffEndDate` is specified and `registrationUitpasNumber` has social tariff, a default date of April 30 of next year is used.",
             "format": "date",
             "writeOnly": true
+          },
+          "registrationPicture": {
+            "type": "string",
+            "description": "Base64 encoded string of the picture of this passholder. Only used in passholder creation.",
+            "format": "byte",
```
bertramakers, line 11918: Is this the correct format for base64-encoded strings? Isn't the point of base64-encoding that the result is text and not bytes?

Or does it have to be `byte` for the library you're using in your codebase?

In any case the `format` is not really standardized anyway, on the one hand that means it doesn't really matter I guess but on the other hand maybe best to avoid it if not strictly needed (to avoid conflicts in client and server libraries)

erwin1:
fixed the conflicts - PR auto-merged now

projects/uitpas/reference/uitpas.json
erwin1: byte means base64 encoded string in OAS 3.0.0 https://spec.openapis.org/oas/v3.0.0.html 

```
byte | string | byte | base64 encoded characters
```

If this adds client-complexity I'm open to drop the format=byte but at least in the generated Java server stub, this nicely maps to a byte[] so I don't have to do the base64 decoding manually.

PR #563, opened 2026-07-17
III-7344/birthdate range move

grubolsch:
### Changed

Moved docs about birthdate from seperate page into existing age page.

Also grouped the age filters into URL parameters and advanced queries, and documented the birthdate range bounds as independently usable (each on its own for an open-ended range, like `dateFrom` and `dateTo`).

> **Note:** The birthdate range docs describe the wanted behaviour, not what is deployed today. The service currently still returns a `404` when only one of `birthdateRangeFrom` / `birthdateRangeTo` is given. Allowing a single bound is implemented in [III-7367](https://jira.publiq.be/browse/III-7367), so this part of the docs leads the code until that ships.

## Live preview

https://publiq.stoplight.io/docs/uitdatabank/branches/III-7344%2FbirthdateRange-move/search-api/filters/filtering-by-target-audience-age#birthdaterangeto

---

Ticket: https://jira.uitdatabank.be/browse/III-7344


grubolsch:
Removed day of week array syntacs

projects/uitdatabank/docs/search-api/filters/age.md
```diff
@@ -2,6 +2,23 @@
 
 It is possible to filter the results from Search API based on the age group the event or place is targetted towards.
 
+## Age filters vs. birthdate range filters
+
+There are two ways to filter on the audience an event or place is targeted towards:
+
+* The `minAge`, `maxAge`, `allAges` and `typicalAgeRange` parameters filter on a **fixed age or age range** (in years), regardless of the current date. They are applicable on `/offers`, `/events` and `/places`.
+* The `birthdateRangeFrom` and `birthdateRangeTo` parameters filter on a **range of birth dates**. They are only applicable on events.
```
lucwollants: I wonder if this is only possible for events. Because, as mentioned later, `typicalAgeRange` is also taken into account?

projects/uitdatabank/reference/search.json
```diff
@@ -1929,6 +1938,29 @@
         "name": "calendarType",
         "description": "Returns only results with the given enum value as their calendarType. Accepts multiple comma-separated values to return results that have one of the given calendar types. [Here is a detailed guide](./entry-api/shared/calendar-info#calendartype) with more information."
       },
+      "dayOfWeek": {
```
lucwollants: I wonder why this `dayOfWeek` change is part of this pull request?

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: Good catch — you are right, and the page contradicted itself on this. The intro said "only applicable on events" while both parameter sections listed `/events` `/offers`, and `search.json` declared them on `/events` alone.

What the code actually does:

- `BirthdateRangeOfferRequestParser` sits in the shared `CompositeOfferRequestParser` (`app/Offer/OfferSearchControllerFactory.php`), so `/offers` and `/places` accept the parameters too — they are not rejected as unsupported.
- `ElasticSearchOfferQueryBuilder::createBirthdateRangeQuery()` ORs a `typicalAgeRange` leg next to the `birthdateRange` leg, which is exactly the point you are making.

Fixed in 3e81a0bc: `/offers` is now declared in the spec as well, so prose and spec agree, and the intro now says what is genuinely event-only — the `birthdateRange` **field**, not the parameters.

One thing worth a separate look, because it may be a bug rather than something to document: since places do carry a `typicalAgeRange`, that fallback leg means a **place** can currently match a birthdate query on `/offers` or `/places`. I deliberately did not document that, and did not add the parameters to `/places`. Either the parser should be scoped to events, or we accept it and say so — your call, but it felt wrong to publish it as intended behaviour.

projects/uitdatabank/reference/search.json
lucwollants: Agreed, it does not belong here — removed in d2a50967.

Besides being unrelated to the birthdate range move (it is III-7246), it could not ship as it stood:

- the `toc.json` entry and the parameter description both linked to `docs/search-api/filters/day-of-week.md`, which exists neither on this branch nor on `main` — two dead links;
- the filter itself is not released, cultuurnet/udb3-search-service#505 and cultuurnet/udb3-backend#2326 are both still open.

The removed content is recoverable from eaef94f2 and 407ad037 whenever it gets its own pull request alongside those two.

projects/uitdatabank/docs/search-api/filters/age.md
```diff
@@ -2,6 +2,23 @@
 
 It is possible to filter the results from Search API based on the age group the event or place is targetted towards.
 
+## Age filters vs. birthdate range filters
+
+There are two ways to filter on the audience an event or place is targeted towards:
+
+* The `minAge`, `maxAge`, `allAges` and `typicalAgeRange` parameters filter on a **fixed age or age range** (in years), regardless of the current date. They are applicable on `/offers`, `/events` and `/places`.
```
bertramakers: Is `typicalAgeRange` an actual parameter? I think it's only a field you can use in advanced queries?

projects/uitdatabank/docs/search-api/filters/age.md
```diff
@@ -2,6 +2,23 @@
...
+* The `birthdateRangeFrom` and `birthdateRangeTo` parameters filter on a **range of birth dates**. They are applicable on `/offers` and `/events`, and always have to be used together. Only events can match: a `birthdateRange` is an event-only field, and places are never returned for it.
+
+<!-- theme: info -->
+
+> An event's `typicalAgeRange` is combined with the current date to determine whether it matches a given `birthdateRange`. Because of this, the same query can return different results on different days.
+
+**When to use which**
+
+* Use the age parameters (`minAge`, `maxAge`, `allAges`, `typicalAgeRange`) when you want events and/or places suitable for a fixed age or age range, for example "events for 6 to 12 year olds".
```
bertramakers: Same question here about `typicalAgeRange`

projects/uitdatabank/docs/search-api/filters/age.md
```diff
@@ -2,6 +2,23 @@
...
+* Use the birthdate range parameters (`birthdateRangeFrom`, `birthdateRangeTo`) when you want events targeted at people born within a specific date range, for example when matching against a known audience of birth dates (e.g. a list of registered children). Since this is date-aware, it also returns events whose `typicalAgeRange` currently overlaps with that birthdate range.
+* Combining both is possible (they are independent parameters and will be applied together) but is only useful for events, since the birthdate range parameters are not supported for places.
```
bertramakers, line 20: In the original ticket I remember it was a requirement that they would be mutually exclusive and you could not use them together. I copied that to the new ticket. But I'm not sure why that was exactly since technically you can use them together indeed.

projects/uitdatabank/docs/search-api/filters/age.md
```diff
@@ -89,3 +106,41 @@ If you don't want partial matches to be returned you can exclude anything that f
 ```http
 GET /events/?q=typicalAgeRange:[6 TO 12] NOT typicalAgeRange:([0 TO 5] OR [13 TO *])
 ```
+
+### birthdateRangeFrom
+
+With the `birthdateRangeFrom` URL parameter you set the earliest birth date of the audience you are looking for. It always has to be combined with `birthdateRangeTo`: supplying only one of the two returns a `404 Not Found`.
```
bertramakers: Is this necessary? Is there a reason why you should not be able to just search on either `birthdateRangeFrom` or `birthdateRangeTo`? How do we handle this on other range parameters like `dateFrom` and `dateTo`?

projects/uitdatabank/docs/search-api/filters/age.md
```diff
@@ -89,3 +106,41 @@ If you don't want partial matches to be returned you can exclude anything that f
...
+
+An event matches when its own `birthdateRange` overlaps the requested range. Events without a `birthdateRange` also match when their `typicalAgeRange`, converted to birth dates using the current date, overlaps the requested range. Events marked as `allAges` are excluded from that fallback, because their age range is unbounded and would otherwise match every birthdate query.
```
bertramakers: Technically an event targeted towards "all ages" is a valid result when look for activities to do with children of a specific age no? Personally I would include events for all ages by default, and allow the integrator to exclude them themselves (either always or if the user chooses to) by setting `?allAges=false` when also searching on birthdateRange and/or typicalAgeRange. That way we support all scenarios in SearchAPI and let the integrator and/or user decide whether events intended for all ages are relevant or not in a specific agenda or search.

projects/uitdatabank/reference/search.json
```diff
@@ -1833,7 +1839,7 @@
         },
         "in": "query",
         "name": "birthdateRangeFrom",
-        "description": "Returns only events targeted at people born on or after the given date. Formatted as one ISO 8601 date."
+        "description": "Sets the earliest birth date of the audience you are looking for. Formatted as one ISO 8601 date. Always has to be combined with `birthdateRangeTo`: supplying only one of the two returns a `404`. Events without a `birthdateRange` also match when their `typicalAgeRange`, converted to birth dates using the current date, overlaps the requested range; events marked as `allAges` are excluded from that fallback."
```
bertramakers: Again not sure why this is required.

projects/uitdatabank/reference/search.json
```diff
@@ -1843,7 +1849,7 @@
         },
         "in": "query",
         "name": "birthdateRangeTo",
-        "description": "Returns only events targeted at people born on or before the given date. Formatted as one ISO 8601 date."
+        "description": "Sets the latest birth date of the audience you are looking for. Formatted as one ISO 8601 date. Always has to be combined with `birthdateRangeFrom`: supplying only one of the two returns a `404`. Matching works the same way as for `birthdateRangeFrom`."
```
bertramakers: Again not sure why this is required.

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: You're right that `typicalAgeRange` is not a URL parameter. The only age URL parameters are `minAge`, `maxAge` and `allAges` (see `AgeRangeOfferRequestParser`). The `typicalAgeRange` can only exists as a field you can use in advanced queries (q=typicalAgeRange:[6 TO 12]).

The parameters are nicely listed in small dedicated sections and are defined as URL parameter or advanced query. So I just removed them from the intro to avoid confusion.

lucwollants: Solved in a6b28ee

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: Solved in a6b28ee

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: You're right that they're not mutually exclusive today. `AgeRangeOfferRequestParser` and `BirthdateRangeOfferRequestParser` are independent parsers and can even be combined. So passing both just applies both, ANDed together as separate FILTER clauses. There is no technical reason they have to exclude each other.

It gets more complicated because `birthdateRange` and `typicalAgeRange` can already be freely combined in an advanced query (q). So if we only make the URL parameters mutually exclusive, we diverge a lot from what q allows anyway.

So I think this is purely a product decision: we decide what we want for the URL parameters, and keep in mind that advanced queries will always allow the combination regardless. Given the flexibility of the advanced query parameters, I would allow combining the URL parameters.

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: Good question. `dateFrom` and `dateTo` can each be used on their own, so you can filter on an open-ended date range. `birthdateRangeFrom` and `birthdateRangeTo` currently require each other and return a `404` when only one is given, so they are not consistent with the other range parameters.

On top of that, you can already do an open-ended range in an advanced query (`q=birthdateRange:[2020-01-01 TO *]`), so requiring both on the URL parameters only diverges from what `q` allows anyway.

So for consistency I think we should allow just one of the two, like `dateFrom`/`dateTo`. I'll create a follow-up ticket for this.

The order matters though, because this depends on [III-7367](https://jira.publiq.be/browse/III-7367). Today the birthdate range is converted to an age range at query time, and that conversion needs both bounds. If we removed the `404` now, we would have to temporarily refactor that query-time conversion to handle a single bound, code that [III-7367](https://jira.publiq.be/browse/III-7367) then throws away again.

Once [III-7367](https://jira.publiq.be/browse/III-7367) indexes both `birthdateRange` and `typicalAgeRange`, that query-time conversion is gone. The parser then just filters on the `birthdateRange` field directly, where a single bound already works out of the box. So doing [III-7367](https://jira.publiq.be/browse/III-7367) first makes this follow-up almost trivial, basically removing the two throws.

projects/uitdatabank/reference/search.json
lucwollants: See https://github.com/cultuurnet/apidocs/pull/563#discussion_r3719980843

projects/uitdatabank/reference/search.json
lucwollants: See https://github.com/cultuurnet/apidocs/pull/563#discussion_r3719980843

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: You want all-ages events included by default, and that made sense with the current query-time conversion. With the index strategy from [III-7367](https://jira.publiq.be/browse/III-7367) it no longer works that way.

The reason: an all-ages event is never converted, so it has no `birthdateRange` in the index (we don't want to invent a fake range). A plain `birthdateRange` query therefore never matches it. All-ages can only be matched through the `allAges` flag, as a separate condition, not through the range.

So all-ages has to be requested explicitly:

- In the `q` parameter the integrator adds it, for example `q=birthdateRange:[2016 TO 2018] OR allAges:true`. We don't touch the `q`, so there is no other way.
- In the URL parameters we have two options: either we add `allAges=true` by default in the query builder, or we don't and we document that integrators add `?allAges=true` when they want all-ages events.

My preference is to keep the URL parameters consistent with the `q`: require an explicit `allAges=true` in both and document it. Then there is no hidden default and both paths behave the same. But I'm fine with default-include on the URL parameters if we think that is friendlier for integrators.

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: No changes are needed to the code, and no changes are needed to the docs. Ticket https://jira.publiq.be/browse/III-7367 will be updated.

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: Remove mention of 404 from the documentation. We want to document what we want to achieve.

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: It should be possible to use null for the boundaries.

lucwollants: Remove "Events marked as `allAges` are excluded from that fallback, because their age range is unbounded and would otherwise match every birthdate query."

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: Sections created in d0a0e6a

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: Removed in e166e65

projects/uitdatabank/docs/search-api/filters/age.md
lucwollants: I removed this in 1bf870a but also made the correlation clear on a higher level and removed the mention of the reference time, which is an implementation detail.

JonasVHG (approved):

bertramakers (approved):

PR #565, opened 2026-07-29
Remove dead link and remove apologetic language

grubolsch:
### Removed

- Removed a link to auth0 that become a dead link, I don't think this a problem because we are not using auth0 anyway.
- Made the warning a little bit less apologetic, I don't think this adds much value/context.


bertramakers (approved):
I'm approving so we can fix the dead link issue asap. But this makes me wonder whether this is still correct because we do use scopes to determine API access, including the new BOA scope. I remember now why it was not a good idea on Auth0 though. I will check with Erwin if it's still an issue with Keycloak.

PR #557, opened 2026-07-14
III-7216 document hasOvernight search filter

grubolsch:
## Summary

* Adds `docs/search-api/filters/overnight.md` documenting the `hasOvernight` boolean filter
* Adds `hasOvernight` to `search.json` (`components/parameters` + referenced in the `/events` GET operation)
* Adds `overnight.md` to `toc.json` so the page is reachable from the sidebar

More BOA audience filters (`hasChildcare`, `childrenOnly`) will follow in subsequent PRs, each as their own page added to `toc.json`.

## Related

* Jira: [III-7216](https://jira.publiq.be/browse/III-7216)
* Backend feature tests: [cultuurnet/udb3-backend#2317](https://github.com/cultuurnet/udb3-backend/pull/2317)

PR #570, opened 2026-08-04
Clarify that subEvent-specific filtering on `childcare` and `overnight` does not work with advanced queries + small fixes

bertramakers:
### Fixed

- Removed factually incorrect reference that `hasChildcare` can be combined with `dateRange` in advanced queries
- Clarified that the combination of `hasChildcare` and `dateFrom`/`dateTo` only works as explicit URL parameters, not in advanced queries since they are parsed differently
- Clarified that the combination of `hasOvernight` and `dateFrom`/`dateTo` only works as explicit URL parameters, not in advanced queries since they are parsed differently
- Removed copy/paste error about `hasOvernight` not influencing the start and end date of events (copied from childcare I assume)

### DIFF
--- a/projects/uitdatabank/docs/search-api/filters/childcare.md
+++ b/projects/uitdatabank/docs/search-api/filters/childcare.md
@@ -37,12 +37,16 @@ GET /events/?hasChildcare=false
 
 ## Combining with a date filter
 
-When `hasChildcare` is combined with a date filter (`dateFrom`/`dateTo` URL parameters, or the `dateRange` advanced query field), the childcare check is scoped to the matching period: the API checks whether childcare is configured on the sub-events or opening hours that fall within that date range, not on the event as a whole.
+When `hasChildcare` is combined with a date filter (`dateFrom`/`dateTo` URL parameters), the childcare check is scoped to the matching period: the API checks whether childcare is configured on the sub-events or opening hours that fall within that date range, not on the event as a whole.
 
 > The date filter itself still matches against the actual start and end times of the activity — childcare hours never influence which events are considered to fall within a date range.
 
-Retrieve all events in May 2025 that offer childcare during that period:
+Retrieve all events in May 2025 that offer childcare during that specific period:
 
 ```http
 GET /events/?hasChildcare=true&dateFrom=2025-05-01T00:00:00%2B02:00&dateTo=2025-05-31T23:59:59%2B02:00
 ```
+
+<!-- theme: warning -->
+
+> While you can also use on childcare and/or dates in advanced queries, this specific logic of checking whether the event has childcare during the specified date range does NOT work in advanced queries (`q` URL parameter) due to how they are parsed differently.
--- a/projects/uitdatabank/docs/search-api/filters/childcare.md
+++ b/projects/uitdatabank/docs/search-api/filters/childcare.md
@@ -49,4 +49,4 @@ GET /events/?hasChildcare=true&dateFrom=2025-05-01T00:00:00%2B02:00&dateTo=2025-
 
 <!-- theme: warning -->
 
-> While you can also use on childcare and/or dates in advanced queries, this specific logic of checking whether the event has childcare during the specified date range does NOT work in advanced queries (`q` URL parameter) due to how they are parsed differently.
+> While you can also filter on `childcare` and/or `dateRange` in advanced queries, this specific logic of checking whether the event has childcare during the specified date range does NOT work in advanced queries (`q` URL parameter) due to how they are parsed differently.
--- a/projects/uitdatabank/docs/search-api/filters/overnight.md
+++ b/projects/uitdatabank/docs/search-api/filters/overnight.md
@@ -2,7 +2,7 @@
 
 ## hasOvernight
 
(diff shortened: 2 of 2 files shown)


JonasVHG (approved):

projects/uitdatabank/docs/search-api/filters/childcare.md
```diff
@@ -37,12 +37,16 @@ GET /events/?hasChildcare=false
...
-Retrieve all events in May 2025 that offer childcare during that period:
+Retrieve all events in May 2025 that offer childcare during that specific period:
 
 ```http
 GET /events/?hasChildcare=true&dateFrom=2025-05-01T00:00:00%2B02:00&dateTo=2025-05-31T23:59:59%2B02:00
 ```
+
+<!-- theme: warning -->
+
+> While you can also filter on `childcare` and/or `dateRange` in advanced queries, this specific logic of checking whether the event has childcare during the specified date range does NOT work in advanced queries (`q` URL parameter) due to how they are parsed differently.
```
lucwollants, line 52: Should this be `hasChildcare` instead of `childcare`?

lucwollants: I think we have the same issue with `overnight`. What is indexed and searchable is `hasOvernight`, so we may need to be more explicit about the naming?

bertramakers: You are correct! But it's annoying because the convention is to use the same names as in the event/place/organizer JSON as much as possible. I don't mind the `has` prefix for booleans but than we should also have used that in the event JSON. I guess the easiest way forward is to just fix it in the docs. But I wonder why Claude didn't complain about this since it's mentioned in the review guidelines in the apidocs reposity:

> New URL query parameters used to filter events, places, offers and organizers must always have the same name as their corresponding field in the event/place/organizer JSON schema.
> ...
> New URL query parameters must have a corresponding Lucene field with the same name that is queryable in the q URL parameter. Make sure this Lucene field is documented in `projects/uitdatabank/docs/search-api/advanced/advanced-queries.md`

Claude should have mentioned that `hasOvernight` and `hasChildcare` are not correct names for the URL parameters and fields in the schema.

Since UiV is not going to use neither of these any time soon, maybe worth fixing? I can create a ticket?

bertramakers: Okay for `hasChildcare` specifically, Claude was not asked to review I see: https://github.com/cultuurnet/apidocs/pull/558
And also not for `hasOvernight`: https://github.com/cultuurnet/apidocs/pull/557
So that explains why it didn't complain 😄

lucwollants: - For `overnight`, it can be a one-on-one mapping between the JSON field `overnight`, which is a boolean, and the indexed field `overnight` instead of `hasOvernight`
- For `childcare`, it is a bit more complex, because the field in JSON is an object with a `start` and `end` date and what we now index is just a `boolean`. Makes me wonder if we should have indexed `childcare` with `start` and `end` and provide a convenience URL parameter `hasChildcare`
- About Claude Code review: maybe it's a good time to trigger it on the creation of the pull request instead of a manual trigger?

bertramakers: - Well spotted! Let's keep `hasChildcare` then since the `childcare` field in the event JSON is not a boolean indeed. The `hasChildcare` is derived, not copied, so it makes sense to give it a separate name 👍 This is consistent with for example the `isDuplicate` field in Elasticsearch which is derived from `duplicateOf`.
- For `overnight` I would personally indeed rename it to remove the `has` then since it's already a boolean in the event JSON and not derived. This is consistent with `childrenOnly`, we didn't rename that to `isChildrenOnly` either. Ideally these fields would have been named `hasOvernight` and `isChildrenOnly` in the UDB3 event JSON to begin with to have a consistent naming for booleans in general in all schemas, but renaming that is going to lead us too far so let's just aim for consistency for each individual field across the APIs.
- Good idea. It will probably increase the cost but good to avoid confusion like this. I will change it for the apidocs repo already.
- Maybe we should look for a way to have some better guidelines for naming? I added  some to the apidocs already before since technically you always need to update those anyway but is that sufficient? And are they complete? Maybe we should add some things like always prefixing new boolean fields with `is` or `has` from the beginning (and consistently across all services)?

lucwollants: For the `hasOvernight` rename to `overnight` I created [III-7191](https://jira.publiq.be/browse/III-7191)

lucwollants: For the Claude Code review automation I created this pull request https://github.com/cultuurnet/apidocs/pull/572

lucwollants: For review guidelines, I created:
- On apidocs https://github.com/cultuurnet/apidocs/pull/573
- On udb3-search-service https://github.com/cultuurnet/udb3-search-service/pull/520

lucwollants (approved):
Indeed! I also forgot about the combination which only works on our URL parameters.

